PR #184, opened 2026-05-05
feat(desktop): header messaging platform status indicators (Phase 4 — U4.1)

huntharo:
First of four stacked PRs for Phase 4 of the design overhaul. Adds per-platform status chips to the right side of the thread header so users can see at a glance which messaging platforms are configured, healthy, and active.

## What lands

**Runtime (main process)**
- \`DesktopMessagingRuntime\` gains a small per-platform status state machine: \`enabled\` / \`suspended\` / \`errored\`, plus per-event activity pings.
- Status survives \`stop()\` so a paused/errored platform stays visible (gray/red dot) — the user knows it's configured, not vanished.
- New emitter API: \`onPlatformStatus(listener)\` + \`getPlatformStatuses()\` snapshot.

**Contract (shared)**
- \`MessagingPlatformHealth\`, \`MessagingPlatformStatus\`, \`MessagingPlatformStatusEvent\` exported from \`@pwragent/shared\`.

**IPC**
- \`messaging:get-platform-statuses\` for initial paint (renderer fetches once).
- \`messaging:platform-status-event\` push bus from main → all renderer windows. Mirrors the agent-event bus pattern.

**Renderer**
- New monochrome \`TelegramIcon\` + \`DiscordIcon\` glyphs filled with \`currentColor\` to match the Tangerine Terminal theme.
- New \`MessagingStatusBar\` renders one chip per configured platform with the \`--status-*\` dot tokens from U0.2. Activity pulses the dot via \`.status-dot--blink\` with a 2s tail; \`prefers-reduced-motion\` disables the animation.
- \`ThreadHeader\` hosts the bar on the right. Renders nothing when no platforms are configured — non-messaging users see the same header.

## Why this shape

The plan's exit condition for U4.1 says:
> Telegram-only configured → only Telegram icon in header. Both configured → both icons. Disabling Telegram in settings → its dot turns gray within 1s. Sending a Telegram message → dot blinks during streaming, stops within ~1s of completion.

The renderer hook \`useMessagingPlatformStatuses\` keeps the canonical \`MessagingPlatformStatus[]\` cache and a separate \`activeAtByPlatform\` map so a flickering activity dot doesn't re-render the rest of the row. The activity tail (2s) is swept by a shared interval rather than per-platform timers.

\`stop()\` marks each previously-running platform as \`suspended\` so the chip stays visible after the user toggles messaging off (U4.2 will exercise this path).

## Tests

5 new runtime tests cover the emission contract:
- \`enabled\` on successful start
[...]

PR #185, opened 2026-05-05
feat(desktop): header messaging on/off toggle (Phase 4 — U4.2)

huntharo:
PR 2 of 4 in Phase 4. Stacks on #184 (U4.1). Adds a global on/off toggle to the right of the per-platform status chips. Persists across launches and respects the startup \`--disable-messaging\` override.

## Runtime

- New \`pause()\` / \`resume()\` on \`DesktopMessagingRuntime\`. \`pause()\` drains via the existing \`stop()\` teardown and emits \`health=suspended\` for each previously-running platform. \`resume()\` re-runs the adapter factory with the *current* config so settings landing during the paused window take effect on resume.
- Both methods are idempotent — multiple \`pause()\`/\`resume()\` cycles are safe.

## Settings

- New \`messaging.userEnabled\` config field (defaults \`true\`). Persists to \`config.toml\` as \`user_enabled\`. Surfaces on the snapshot under \`runtime.messaging.userEnabled\` so the renderer toggle binds to it.
- Bootstrap reads the snapshot and skips \`runtime.start()\` when \`userEnabled\` is \`false\`.

## IPC

- New \`messaging:set-enabled\` handler. When the startup override is in force, returns \`{ overridden: true }\` and refuses to mutate state. Otherwise persists the setting and calls \`runtime.pause()\`/\`resume()\`.
- Settings write fires *before* the runtime transition so a crash mid-toggle leaves a consistent on-disk truth.

## UI

- \`MessagingStatusBar\` gains a small pill toggle on the right with optimistic state and rollback on IPC failure.
- When the CLI/env override is in force, the toggle renders disabled with a tooltip explaining which override fired.

## Tests

- 3 new runtime tests for pause/resume + idempotent cycle.
- Snapshot fixtures updated across renderer + service tests for the new \`userEnabled\` field.

Total: 1129 passing / 3 skipped.

## Stacking

- This is PR 2 of 4. Stacks on [#184](https://github.com/pwrdrvr/PwrAgent/pull/184) (U4.1).
- Next: **U4.3** — per-thread binding chips + unbind menu (independent of U4.2; will stack on U4.1 directly).
- After: **U4.4** — Messaging Activity screen (stacks on U4.3).

## Post-Deploy Monitoring & Validation

- **What to monitor/search**
  - Logs: \`pwragent:main\` for \`messaging runtime not started — disabled by user setting\` on launch (should fire for users who toggled off).
  - Logs: \`pwragent:messaging\` for \`messaging runtime started\` on resume (should fire when the user toggles back on).
- **Validation checks**
  - Toggle off → adapters stop within ~1s; chips persist with gray dots.
  - Toggle on → adapters re-init and report green within 2s.
  - Quit + relaunch → toggle state remembered.
  - Launch with \`--disable-messaging\` → toggle is disabled with the tooltip explaining the flag.
- **Expected healthy behavior**
  - No flicker mid-toggle; pill animation is smooth.
  - Settings file gains \`messaging.user_enabled = true|false\` after first toggle.
- **Failure signal(s)**
  - Toggle gets stuck in optimistic state → IPC handler rejected without rollback (look for unhandled exceptions in renderer devtools).
  - \`adapter.stop is not a function\` or hangs → adapter author missing the \`stop()\` impl; would surface as a stuck \"Off\".
- **Validation window & owner**
  - Window: First post-deploy launch + 5 min of toggle exercise.
  - Owner: @huntharo

[Compound Engineering v2.49.0](https://github.com/EveryInc/compound-engineering-plugin)
🤖 Generated with [Claude Opus 4.7 (1M context, extended thinking)](https://claude.com/claude-code) via Compound Engineering v2.49.0

### DIFF
--- a/apps/desktop/src/main/__tests__/desktop-settings-service.test.ts
+++ b/apps/desktop/src/main/__tests__/desktop-settings-service.test.ts
@@ -354,6 +354,7 @@ describe("DesktopSettingsService", () => {
     expect(snapshot.runtime.messaging).toEqual({
       disabled: true,
       disabledReason: "--disable-messaging was provided at startup",
+      userEnabled: true,
     });
   });
 
--- a/apps/desktop/src/main/__tests__/index.test.ts
+++ b/apps/desktop/src/main/__tests__/index.test.ts
@@ -28,6 +28,11 @@ const messagingRuntimeStartMock = vi.fn<() => Promise<void>>();
 const disposeDesktopMessagingRuntimeMock = vi.fn();
 const registerMessagingStatusIpcHandlersMock = vi.fn();
 const disposeMessagingStatusIpcHandlersMock = vi.fn();
+const settingsReadSettingsMock = vi.fn(async () => ({
+  fetchedAt: 0,
+  configPath: "/tmp",
+  runtime: { messaging: { disabled: false, userEnabled: true } },
+}));
 const setApplicationMenuMock = vi.fn();
 const buildFromTemplateMock = vi.fn(() => ({ kind: "menu" }));
 const setNameMock = vi.fn();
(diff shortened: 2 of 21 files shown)


huntharo:
Folded into #188 (single PR landed against main).

PR #213, opened 2026-05-06
fix(desktop): send approvalPolicy and sandboxPolicy on every codex turn/start

huntharo:
## Summary

After toggling Default → Full → Default on a thread, the next turn ran without a sandbox even though every PwrAgent surface labeled the thread as Default Access. The overlay flipped correctly, routing picked the default codex client process correctly (verified by the diagnostic logs from #203 and the spawn-args fix from #209) — but the codex thread session inside that process still had the **Full Access PermissionProfile** baked into it from the prior turn.

## Why this was missing

Codex underwent a big refactor (#18278, #19773–19776, #20106 upstream) that moved permission state from process-level to per-thread `PermissionProfile`. The process-level `-c sandbox_mode=...` flags now only seed the *default* profile for **newly created** threads — they don't govern existing threads. Once a thread has a profile, the only way to change it is to send fresh `approvalPolicy`/`sandboxPolicy` overrides on `turn/start` (or via `thread/resume`).

Codex's V2 `TurnStartParams` documents both fields as overrides "**for this turn and subsequent turns**" — this is the protocol-blessed way to refresh the per-thread profile each turn. The codex TUI sends both on every `turn/start` ([`app_server_session.rs:540-565`](https://github.com/openai/codex)).

Our desktop was dropping both at the protocol boundary:
- [`buildTurnStartPayload`](apps/desktop/src/main/codex-app-server/client.ts:3110) didn't accept `approvalPolicy` or `sandbox` as parameters.
- [`client.startTurn`](apps/desktop/src/main/codex-app-server/client.ts:3855) called it without those fields.

The backend-registry layer was already computing the right values from `EXECUTION_MODE_SUMMARIES` and passing them down — they were silently discarded before hitting the wire.

## What changed

- `buildTurnStartPayload` now accepts `approvalPolicy?: string` and `sandbox?: string` and forwards them in the V2 wire shape.
- `client.startTurn` forwards both from the registry.
- New helper `buildCodexSandboxPolicy` converts the legacy `SandboxMode` string to the structured `SandboxPolicy` object that V2 `turn/start` expects:
  - `"workspace-write"` → `{ type: "workspaceWrite", writableRoots: [], networkAccess: false, excludeTmpdirEnvVar: false, excludeSlashTmp: false }`
  - `"danger-full-access"` → `{ type: "dangerFullAccess" }`
  - `"read-only"` → `{ type: "readOnly", networkAccess: false }`

`thread/resume` (called before each `turn/start` and from `setThreadPermissions` for direct toggles) still uses the legacy `sandbox: SandboxMode` field — codex accepts that for backward compat (the upstream commit "permissions: store thread sessions as profiles" explicitly preserved it as a compatibility boundary).

## Stack with previous fixes

This completes the chain:
- #203: Composer was dropping `executionMode`, registry had a silent cross-mode fallback.
- #209: default codex client was spawned with no `-c` args, inheriting permissive upstream defaults.
- This PR: per-thread `PermissionProfile` was never refreshed on toggle, so a previously-full-access thread stayed full-access.

Each fix is a different layer of the same root cause (permission state migrated from process to thread). Together, every layer now agrees with the user's chosen mode.

## Tests

Adds two regression tests in [`codex-client.test.ts`](apps/desktop/src/main/__tests__/codex-client.test.ts) asserting the exact wire shape:

- `workspace-write` + `on-request` → `sandboxPolicy: { type: "workspaceWrite", … }`, `approvalPolicy: "on-request"`
- `danger-full-access` + `never` → `sandboxPolicy: { type: "dangerFullAccess" }`, `approvalPolicy: "never"`

Future drift in either the policy fields or the conversion logic fails CI.

## Test plan

- [x] `npx vitest run apps/desktop/src/main/__tests__/codex-client.test.ts` — 56/56 pass (54 existing + 2 new).
- [x] `npx vitest run --config vitest.workspace.ts apps/desktop/src/main/__tests__/` — 609/609 desktop-main pass.
- [x] `npx vitest run --config vitest.workspace.ts apps/desktop/src/renderer/` — 348/348 renderer pass.
- [x] `pnpm --filter @pwragent/desktop exec tsc --noEmit` — clean.
- [ ] User repro: toggle Default → Full → Default on a real thread, send a network command (e.g. `npm view express`), confirm approval prompt fires.

### DIFF
--- a/apps/desktop/src/main/__tests__/codex-client.test.ts
+++ b/apps/desktop/src/main/__tests__/codex-client.test.ts
@@ -66,6 +66,9 @@ class MockTransport implements JsonRpcTransport {
     rateLimitsByLimitId: {}
   };
   static turnInterruptResponseMode: "success" | "timeout" = "success";
+  static threadResumeError:
+    | { code?: number; message: string }
+    | undefined = undefined;
 
   readonly sentMessages: string[] = [];
   private messageHandler: (message: string) => void = () => undefined;
@@ -641,6 +644,19 @@ class MockTransport implements JsonRpcTransport {
     }
 
     if (payload.method === "thread/resume") {
+      if (MockTransport.threadResumeError) {
+        this.messageHandler(
+          JSON.stringify({
+            jsonrpc: "2.0",
+            id: payload.id,
+            error: {
+              code: MockTransport.threadResumeError.code ?? -32000,
+              message: MockTransport.threadResumeError.message,
+            },
+          })
+        );
+        return;
+      }
       this.messageHandler(
         JSON.stringify({
           jsonrpc: "2.0",
@@ -835,6 +851,7 @@ describe("CodexAppServerClient", () => {
       rateLimitsByLimitId: {}
     };
     MockTransport.turnInterruptResponseMode = "success";
+    MockTransport.threadResumeError = undefined;
   });
 
   it("initializes once and normalizes thread/list results", async () => {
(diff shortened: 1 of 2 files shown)


huntharo:
Superseded by #219, which absorbs this PR's commit (`e9d6be43` — per-turn `approvalPolicy`/`sandboxPolicy` on `turn/start` + resume-failure logging) into the larger single-instance + queued-permissions architecture. The per-turn override is no longer a transitional defense-in-depth; it's part of the stable architecture documented in `docs/solutions/2026-05-07-codex-permission-mode-state-machine.md`. Closing without merging — the changes ship via squash-merge on #219.

PR #329, opened 2026-05-11
fix(telegram): align General topic typing target

huntharo:
## Summary

- I fixed Telegram General topic typing so the typing indicator uses the same unthreaded target as General-topic message sends.
- This avoids renewing typing on `message_thread_id: 1` and then delivering the assistant reply to the broader chat target, which can leave Telegram showing stale typing after the reply appears.
- I updated the regression coverage to assert General-topic typing and message sends both omit `message_thread_id`.

## Verification

- `pnpm test apps/desktop/src/main/__tests__/telegram-adapter.test.ts -t "uses unthreaded General topic routing"`
- `pnpm test apps/desktop/src/main/__tests__/telegram-adapter.test.ts`
- `pnpm test apps/desktop/src/main/__tests__/messaging-controller.test.ts -t "routing state|typing"`
- `pnpm test`
- `pnpm typecheck`
- `pnpm lint:boundaries`

## Notes

- PR #326 was already merged, so this is a follow-up branch on top of current `origin/main`.

### DIFF
--- a/apps/desktop/src/main/__tests__/telegram-adapter.test.ts
+++ b/apps/desktop/src/main/__tests__/telegram-adapter.test.ts
@@ -504,7 +504,7 @@ describe("TelegramAdapter", () => {
     expect(api.sendMessage).not.toHaveBeenCalled();
   });
 
-  it("uses General topic routing state for typing without threading message sends", async () => {
+  it("uses unthreaded General topic routing for typing and message sends", async () => {
     const api = createApi();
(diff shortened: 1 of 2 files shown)


huntharo:
Closing this follow-up: live testing showed General-topic typing needs message_thread_id=1 to display in Telegram. The branch now reverts the proposed normalization, so there is no useful diff to merge. The original stuck-typing incident was likely caused by another running bot instance renewing typing.

PR #351, opened 2026-05-12
feat(desktop): add profile messaging leases

huntharo:
## Summary

- Added profile-scoped runtime instance logging and a singleton messaging lease in the desktop profile SQLite DB.
- Routed desktop startup, settings runtime toggles, config hot-apply, and pairing approval through a lease coordinator so only one app instance per profile starts messaging adapters.
- Updated Settings runtime status, shared contracts, and developer docs so a second instance reports "lease held elsewhere" instead of requiring `dev:no-messaging`.

## Verification

- I ran `pnpm test`.
- I ran `pnpm lint`.
- I ran `pnpm -r --if-present typecheck`.
- I ran `git diff --check`.

## Notes

- `dev:no-messaging` remains available as an explicit session opt-out; normal `pnpm dev` now records the instance and only starts adapters after acquiring the profile lease.
- Lease takeover is passive: another instance can acquire messaging after the prior holder stops heartbeating and the lease expires.

### DIFF
--- a/apps/desktop/AGENTS.md
+++ b/apps/desktop/AGENTS.md
@@ -41,13 +41,13 @@ The desktop style guide defines:
 To launch the desktop app with live threads and real user state, run from the **repo root** (or worktree root):
 
 ```bash
-pnpm --filter @pwragent/desktop dev:no-messaging
+pnpm --filter @pwragent/desktop dev
 ```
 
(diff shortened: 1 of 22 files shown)


huntharo:
Superseded by #350. The profile messaging lease commits from this PR were cherry-picked into #350 and the launcher work now builds on them there.

huntharo:
Closing as superseded by #350.

Issue #385, opened 2026-05-13
feat(messaging): render inbound PDFs to images before model upload


huntharo:
## Problem

Messaging-platform PDF uploads are currently forwarded to the model as native PDF file input. That avoids PwrAgent doing brittle text extraction, but in practice the model often burns local tool calls and tokens converting the PDF into page images anyway before it can answer visual/document-layout questions.

That makes the workflow slower and more expensive, and it can surprise users because a simple "what's in this PDF?" request may trigger Python installs, virtualenv creation, PDF rendering, or other local conversion work inside the agent turn.

## Proposal

Add first-class PDF page rendering in PwrAgent's messaging attachment pipeline:

- Detect accepted inbound PDF attachments after the existing size/count checks.
- Render selected PDF pages to normalized image input before starting the model turn.
- Upload those images instead of, or alongside, the PDF depending on a setting.
- Preserve the current native-PDF pass-through path as an option.
- Add a user-facing setting for the PDF handling mode, for example:
  - `native_pdf`: send the PDF as a file input.
  - `rendered_images`: render pages to image input and do not send the PDF.
  - `auto`: choose based on page count, provider/model capability, or attachment size.
- Add safety controls for max pages, image quality/profile, rendered pixel budget, and failure fallback.
- Keep rendered image payloads in memory only, matching the current attachment privacy/persistence behavior.

The default probably wants to be conservative at first, but the setting should let users choose the token/tool-call tradeoff explicitly.

## Alternatives

- Keep native PDF pass-through only and let the model decide when to render pages. This is simple but pushes avoidable work into agent turns.
- Send both PDF and rendered images every time. This maximizes model context, but risks wasting tokens for text-heavy PDFs and duplicate information.
- Go back to text extraction for PDFs. That loses layout/image-only content and was the behavior we intentionally moved away from.

huntharo:
Completed in an evolved form by PR #691. PwrAgent now performs bounded local PDF analysis: one-page PDFs become model images, multi-page PDFs use bounded initial rendering plus authenticated page-render/search tools, limits cover pages/pixels/bytes/model input, and a default-on local-PDF-analysis setting preserves the reference-only fallback. This supersedes the original eager render-all-pages/native-PDF mode proposal while solving the same tool-call and privacy problem.

PR #386, opened 2026-05-13
fix(messaging): pass through PDF attachments

huntharo:
## Summary

- I changed messaging PDF ingestion to pass PDFs through as native file input instead of extracting PDF text.
- I added shared file input/message-part contracts and wired Codex, Grok, replay persistence, and transcript rendering around file metadata.
- I updated the messaging docs and added focused coverage for PDF pass-through, AI SDK file parts, controller routing, and replay sanitization.

## Verification

- I ran `pnpm test apps/desktop/src/main/__tests__/messaging-attachment-processor.test.ts apps/desktop/src/main/__tests__/messaging-controller.test.ts packages/agent-core/src/__tests__/ai-sdk-message-builder.test.ts packages/agent-core/src/__tests__/session-state.test.ts`.
- I ran `pnpm typecheck`.
- I ran `pnpm lint:boundaries`.

## Notes

- This keeps PDF bytes in turn input only; replay/transcript persistence stores file name, MIME type, and size metadata, not the base64 PDF payload.
- Follow-up issue for optional PDF-to-image preprocessing: #385.


huntharo:
Addressed the Codex protocol concern from review. I checked the installed Codex Desktop bundle: non-image files are treated as local file references, with prompt context listing the saved paths, rather than raw file UserInput variants. This branch now mirrors that at the Codex adapter boundary: file inputs are written to a temp path and sent as text context + a mention, so turn/start never includes an unsupported type: file item. Grok still receives native file inputs.

PR #588, opened 2026-05-29
fix(desktop): restore clickability for onboarding and warning overlay

serhiikushch:
## Summary
In the onboarding wizard, most controls were not clickable while the overlay was open (notably only `Skip setup` remained usable).
`codex-config-warning-banner` controls could become non-interactive in overlapping drag-region contexts.

## Changes
- Fixed onboarding wizard overlay clickability regression without breaking layout:
  - Kept `.onboarding-wizard-overlay` fixed/centered.
  - Set overlay to `-webkit-app-region: no-drag`.
  - Kept scrim non-interactive (`pointer-events: none`) and behind wizard.
  - Raised wizard above scrim with explicit z-index.
- Restored window dragging while wizard is open:
  - Made `.onboarding-wizard__titlebar` a drag region.
  - Marked interactive titlebar controls as `-webkit-app-region: no-drag`.
- Fixed identical clickability issue for `codex-config-warning-banner` by setting it to `-webkit-app-region: no-drag`.
- Added renderer style contract coverage for this drag/click behavior.
- Added a focused onboarding E2E regression test that validates `Get started` and `← Back` click transitions.

## Verification

- `pnpm test apps/desktop/src/renderer/src/styles/__tests__/theme-contract.test.tsx` ✅
- `pnpm --filter @pwragent/desktop test:e2e e2e/onboarding-wizard.spec.ts --grep "Get started and Back buttons are clickable"` ✅ 
- manual test: onboarding completed ✅

### DIFF
--- a/apps/desktop/e2e/onboarding-wizard.spec.ts
+++ b/apps/desktop/e2e/onboarding-wizard.spec.ts
@@ -26,6 +26,39 @@ const wizardLaunchOptions = {
 };
 
 test.describe("Onboarding wizard", () => {
+  test("Get started and Back buttons are clickable", async () => {
+    const app = await launchElectronApp(wizardLaunchOptions);
+    try {
+      await expect(
+        app.window.getByRole("heading", {
+          name: /A few short choices/i,
+        }),
+      ).toBeVisible();
+
+      const getStarted = app.window.getByRole("button", { name: /Get started/i });
+      await expect(getStarted).toBeVisible();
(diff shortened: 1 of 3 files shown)


huntharo:
@serhiikushch - Thanks for this!  I was able to reproduce and the fix works well!  I didn't notice this because it doesn't happen if re-invoking the wizard but would happen to EVERY new user.

PR #627, opened 2026-06-02
fix(desktop): add sidebar masthead action tooltips

serejja:
## Summary
- add visible hover and focus tooltips to the sidebar masthead Automations, Settings, and New Thread icon buttons
- reuse the existing viewport-safe sidebar tooltip pattern so captions are not clipped and existing button behavior stays intact
- add sidebar regression coverage that verifies the new captions and preserves the original click handlers

## Testing
- `pnpm test apps/desktop/src/renderer/src/features/navigation/__tests__/sidebar.test.tsx`

## Post-Deploy Monitoring & Validation
No additional operational monitoring required. This is a renderer-only tooltip affordance change with no backend, persistence, or runtime contract impact.


serejja:
https://github.com/user-attachments/assets/b8f6056b-8861-4e12-af60-624b79f5b6bf

PR #637, opened 2026-06-03
fix(desktop): approve notification requests from native alerts

serejja:
## Summary
- add an `Approve` action to native desktop approval notifications on supported platforms
- resolve the existing pending server request from the notification path so approval clears the pending state through the normal `serverRequest/resolved` flow
- move pending-request response mapping into `@pwragent/shared` so the renderer and notification approval path use the same decision translation

## Testing
- `pnpm test apps/desktop/src/main/__tests__/desktop-notification-service.test.ts`
- `pnpm test apps/desktop/src/main/__tests__/backend-registry.test.ts --testNamePattern "resolves a pending approval when the notification approve action fires|calls notifyAttention on turn/requestApproval with a backend:thread:request key"`
- manual replay-backed desktop verification on macOS: triggered a native approval notification and confirmed the notification `Approve` button resolves the pending approval

## Notes
- unsupported platforms keep the passive notification behavior with no action button
- macOS still focuses the app when the notification action is pressed; this PR only makes the action approve successfully

--------
[Compound Engineering](https://compound.engineering)

### DIFF
--- a/apps/desktop/electron-builder.yml
+++ b/apps/desktop/electron-builder.yml
@@ -98,6 +98,7 @@ mac:
   helperPluginBundleId: com.pwrdrvr.pwragent.helper.Plugin
   extendInfo:
     NSHumanReadableCopyright: "Copyright © 2026 PwrDrvr LLC."
+    NSUserNotificationAlertStyle: alert
     LSMinimumSystemVersion: "12.0"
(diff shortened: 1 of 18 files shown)


serejja:
https://github.com/user-attachments/assets/4d25f5dc-80bc-418a-a338-388a64bfdadb

PR #758, opened 2026-06-11
fix(desktop): keep task monitor commands child-owned

huntharo:
## Summary
Task monitors now steer Codex away from the failure mode where a parent starts a long-running command and delegates only the parent-local exec session id. The monitor prompt and parent guidance now make the intended ownership explicit: delegate before running, let the monitor child run the command in its own Codex tool session, poll its own stdout/stderr buffers, and include captured output file paths in the final handoff.

PwrAgent now rejects monitor delegations that reference parent-scoped `exec session` / `write_stdin session_id` handles, so the parent gets an immediate actionable error instead of spawning a monitor that cannot attach to the session and returns no test output.

## Tests
- `pnpm --filter @pwragent/desktop exec vitest run src/main/__tests__/backend-registry.test.ts -t "task monitor"`

---

GPT--5


huntharo:
Superseded by #759

PR #867, opened 2026-06-18
fix(desktop): avoid unnecessary keychain prompts

huntharo:
## Summary
PwrAgent no longer asks macOS Keychain for the “PwrAgent Safe Storage” item just because a stored Grok API key exists. Settings snapshots now report whether secrets are configured without decrypting them, and the direct AgentCore-Grok backend only resolves the xAI key after that experimental feature is enabled.

This keeps normal startup, shell environment hydration, and settings metadata reads from surfacing a scary password prompt. The Grok key is still decrypted when a feature that actually needs it runs.

## Tests
- `pnpm test apps/desktop/src/main/__tests__/desktop-settings-service.test.ts apps/desktop/src/main/__tests__/backend-registry.test.ts`
- `pnpm lint:boundaries`

---

GPT-5

### DIFF
--- a/apps/desktop/src/main/__tests__/backend-registry.test.ts
+++ b/apps/desktop/src/main/__tests__/backend-registry.test.ts
@@ -1588,6 +1588,32 @@ describe("DesktopBackendRegistry", () => {
     }
   });
 
+  it("does not resolve the Grok API key while AgentCore-Grok is disabled", async () => {
+    const previous = process.env.PWRAGENT_EXPERIMENTAL_AGENT_CORE_GROK;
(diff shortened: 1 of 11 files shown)


huntharo:
> When a safeStorage row exists but can no longer be decrypted ... this metadata-only check makes settings snapshots report `configured: true`/`source: "keychain"`.

Addressed in `179cb936`: the DB-backed secret store now records decrypt failures in-process and reports those rows as unset for later metadata reads, while clearing the marker when the secret is replaced or deleted. This preserves the no-decrypt settings snapshot path but prevents a known-unusable encrypted row from continuing to appear configured.

PR #886, opened 2026-06-24
fix(desktop): avoid sqlite autoincrement churn

huntharo:
## Summary

PwrAgent no longer keeps SQLite sequence counters for bounded append tables whose retained rows are capped. Existing profile databases migrate `messaging_activity_log` and `composer_draft_journal` to ordinary integer rowids, preserving rows and indexes while clearing stale `sqlite_sequence` entries so insert/prune churn cannot be hidden behind small retained row counts.

## Validation

- `pnpm test apps/desktop/src/main/__tests__/state-db.test.ts`
- `pnpm lint`

---

GPT-5

### DIFF
--- a/apps/desktop/src/main/__tests__/state-db.test.ts
+++ b/apps/desktop/src/main/__tests__/state-db.test.ts
@@ -80,101 +80,6 @@ describe("StateDb", () => {
     ]);
   });
 
-  it("creates bounded append tables without AUTOINCREMENT counters", () => {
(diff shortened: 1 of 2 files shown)


huntharo:
Closing after re-evaluating the migration risk: the branch now reverts the AUTOINCREMENT removal, leaving no file diff. We should keep the existing SQLite schema and use a narrower diagnostic/source-level mitigation if needed.

PR #943, opened 2026-07-05
refactor(desktop): replace the 1M-token pricing summary heuristic with an explicit scope

huntharo:
## Why

`isHistoricalUsageSummary` classified a pricing usage line as a whole-thread summary with a magic-number heuristic:

```ts
line.source === "live" && line.status === "pending" &&
line.cumulativeTotalTokens === undefined && line.totalTokens >= 1_000_000
```

Two problems:
1. **It re-guesses on every render.** The classification is derived from raw token counts each paint, forever.
2. **The threshold is aging badly.** It was safe when turns rarely crossed 1M tokens, but large cached contexts routinely push a *single legitimate turn* past 1M now — so an active turn could get labeled "Historical usage summary" (and, in the durations work, wear a Live chip with no clock).

The rows the heuristic actually exists for are **legacy**: they were persisted before the Codex protocol carried per-turn cumulative snapshots, so a whole-thread total landed as one `live`/`pending` row scoped as a `turn`. That's a **finite, static set already sitting in the sqlite ledger** — not something new writes produce (modern live rows carry `cumulative_total_tokens`).

## What changed

Fix the data once instead of guessing on every render:

- **`state-db` migration (`user_version` 24)** — `reclassifyLegacyThreadUsageSummaries` runs a one-shot `UPDATE` marking legacy rows (`source='live' AND status='pending' AND scope='turn' AND cumulative_total_tokens IS NULL AND total_tokens >= 1000000`) to `scope = 'total'`, the structural "historical summary" scope that already routes through the principled first branch. Modern turns (with cumulative) and sub-1M turns are untouched. The token threshold now runs **once, against a closed legacy set** — not on every paint.
- **`isHistoricalUsageSummary`** drops to `scope === "total" || scope === "backfill"`. No token arithmetic. A large live turn is now correctly a **Turn** (with its per-turn cost suffix), not a summary.

Net effect on display is unchanged for the legacy rows (still "Historical usage summary", still `$X list price` not `list price this turn`), because `scope: 'total'` produces exactly that. The behavior *change* is the intended one: a genuinely large live turn is no longer misclassified.

Reader for the constant-bump gotcha: the previous last migration step set `user_version` to `CURRENT_STATE_DB_USER_VERSION`; that's now pinned to a literal `23` so bumping the constant to 24 can't skip the new step.

## Testing

- New main-process test: migration reclassifies a legacy live/pending 2M-token summary row to `scope 'total'` while leaving a cumulative-bearing modern turn and a sub-1M live turn as `'turn'`.
- Renderer tests: a `scope: 'total'` row renders "Historical usage summary" / `$X list price`; a large `scope: 'turn'` live row now renders "Turn usage" / `$X list price this turn`.
- `typecheck`, `lint:eslint`, `lint:sql`, `lint:boundaries`, and the `state-db` / `state-migration` suites all pass.

## Relationship to #935

Independent and stacks cleanly. #935 (per-turn durations) added a guard-order fix so an active turn always shows its clock even when the old heuristic misfired; this PR removes the misfire at its root. Whichever merges second reconciles trivially — they touch different functions.

### DIFF
--- a/apps/desktop/src/main/__tests__/overlay-store-usage-pricing.test.ts
+++ b/apps/desktop/src/main/__tests__/overlay-store-usage-pricing.test.ts
@@ -492,6 +492,61 @@ describe("SqliteOverlayStore thread usage pricing ledger", () => {
       }),
     ]);
   });
+
+  it("reclassifies legacy live summary rows to scope 'total' on migration", async () => {
+    // A pre-cumulative-snapshot whole-thread total that landed as a live/pending
+    // "turn" row. It should be reclassified; a modern turn (has cumulative) and a
+    // sub-1M live turn should not.
+    await store.upsertThreadUsageLine({
+      line: buildUsageLine({
+        scope: "turn",
+        source: "live",
+        status: "pending",
+        totalTokens: 2_000_000,
+        turnId: "turn-legacy",
+        usageLineId: "legacy-summary",
+      }),
+    });
+    await store.upsertThreadUsageLine({
+      line: buildUsageLine({
+        cumulativeTotalTokens: 2_000_000,
+        scope: "turn",
+        source: "live",
+        status: "pending",
+        totalTokens: 2_000_000,
+        turnId: "turn-modern",
+        usageLineId: "modern-turn",
+      }),
+    });
+    await store.upsertThreadUsageLine({
+      line: buildUsageLine({
+        scope: "turn",
+        source: "live",
+        status: "pending",
+        totalTokens: 500_000,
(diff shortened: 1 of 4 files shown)


huntharo:
Closing in favor of a corrected approach. This PR classified usage rows via a one-time migration + kept the renderer doing the classification; the right fix is to decide attribution upstream (where deriveLiveThreadTokenUsage knows whether it could attribute usage to a turn) and persist an explicit marker, so the renderer stops guessing. New work will stack on #935.

PR #992, opened 2026-07-14
fix(desktop): stop orphaning integrated terminals when the thread view unmounts

huntharo:
## The bug

Terminals detach from the UI. Ten shells were running in my instance with nothing on screen pointing at them — the only evidence was the quit dialog's blocker count.

**The main process already owned the PTYs.** `IntegratedTerminalService` already spawned them, held them in `sessionsByThread` / `sessionsById`, and killed them on shutdown. Nothing about that moved.

What was missing is the **read half of the IPC**. The channels were `create | write | resize | close | output | exit | error` — all writes and pushes. There was no way for the renderer to *ask* "what is running?" So the renderer kept its own authoritative copy of "which terminals exist and which panel is open" in two `useState` maps inside `ThreadView` ([`terminalOpenByThread`, `retainedTerminalByThread`](apps/desktop/src/renderer/src/features/thread-detail/ThreadView.tsx)) — and any unmount wiped all of it at once while main happily kept the shells running.

`ThreadView` unmounts on ordinary navigation:
- opening the **search screen** (`mainView === "search"`), and
- any moment **`threadDetailPending`** flips — the selected thread transiently missing during a refresh.

The overnight case is the second. `useThreadNavigation` treats >30 min of *wall-clock* idle as idle, so a laptop suspend qualifies; the forced `active-recent` refresh on wake returns a reduced snapshot that can drop the selected thread just long enough to unmount.

Verified against the live instance: exactly ten `/bin/zsh -l` children of the main process, matching the dialog's count, with nothing in the UI referencing any of them.

## The fix

Main publishes its registry, and becomes the source of truth **for the UI**:

- Four new channels — `integrated-terminal:list`, `:sessions` (broadcast), `:set-panel-hidden`, `:reveal`.
- `useIntegratedTerminals` is instantiated in `App` — above every unmount point — mirrors the registry, and reattaches on remount (`createOrAttach` already replayed the scrollback buffer; it just had no caller).
- `panelHidden` moves into the session record, so "I collapsed this" outlives the component that rendered it.

**A collapsed-but-running terminal now advertises itself**: the thread-header toggle wears a live dot and reads "Show running integrated terminal"; sidebar thread rows get a terminal chip. If you never collapsed it, the panel simply comes back.

## Quit dialog

Lists the work instead of counting it — scrollable, grouped (turns / terminals / environment actions), each row a link that cancels the quit and navigates to the thread, revealing its terminal panel for terminal rows.

The countdown scales with row count (ten terminals is not a ten-second read) and any deliberate interaction cancels it — notably **not** `focusin` or `pointermove`, since the Quit button is autofocused and a pointermove fires when the window merely appears under a stationary cursor. Both cancelled it before anyone had read a word, which would have left an unattended shutdown waiting forever for a human who wasn't there.

The dialog also gains a `default-src 'none'` CSP and denies non-signal navigation: it now interpolates model-generated thread titles and user command strings, so an escaping slip must not become exfiltration.

## Environment actions were leaking

They were invisible to the quit path entirely, **and `disposeMainProcessResourcesSync` never killed their detached process trees**. They were abandoned on quit and only *usually* died of SIGPIPE — a quiet `pnpm dev` could outlive the app holding its port. Now counted, listed, and actually stopped.

Subtler: only the Run-button path had an owner, so an action started by *applying an environment* never reached the blockers while still being hard-killed. If it was the only work running, the blocker count was zero and quit took the "no active work" fast path — no dialog at all. The launchpad spawns its action *before* the thread exists, so the owner is registered without a thread id and back-filled once `startThread` returns.

## Thread-key canonicalization

`formatQuitThreadKey` and ThreadView's terminal key both hand-rolled `${backend}:${threadId}`, which is ambiguous for ACP backends whose kind (`acp:grok`) already contains a colon. Both now use `buildThreadIdentityKey`. This was a latent bug for ACP threads.

## Diagnostics

- **Copyable process IDs** in Settings → About and Settings → Developer mode. Main owns the terminal registry, so main's pid is the useful one — and with several Electron apps running you couldn't tell which was PwrAgent.
- **On-demand heap snapshots** (immediate / 5s / 10s) of both processes. The existing heap capture only fired off a CPU spike, which is exactly wrong for an idle process holding onto something it shouldn't. Gated on Developer Mode, and every surface warns that a main-process snapshot contains decrypted API keys and bot tokens — do not hand the path to a bug report or an agent.

## Verification

- [`recovers a running terminal after the thread view unmounts`](apps/desktop/e2e/integrated-terminal.spec.ts) and [`keeps a collapsed terminal collapsed across a thread view unmount`](apps/desktop/e2e/integrated-terminal.spec.ts) drive a **real PTY** through the exact bug paths. Both fail on the pre-fix code.
- [`quit-confirmation-dialog.spec.ts`](apps/desktop/e2e/quit-confirmation-dialog.spec.ts) — the dialog had **no** E2E at all, and a `data:` page that signals main by navigating a fake scheme fails silently while still looking perfect. It caught the autofocus regression above.
- 4629 unit tests, 364 files. Desktop E2E green in CI.


huntharo:
## Review fixes pushed (d1ce0a5c)

All seven findings addressed, plus one more the new E2E caught. Three were behavior this PR claimed to have and didn't.

### Blockers

**Collapsed terminals popped back open on remount** — and not just the one you navigated to. `createOrAttach` un-hid the panel on every attach, on the theory that the renderer only attaches when it's showing the pane. It isn't: `ThreadView` mounts a pane (and therefore attaches) for *every* live session, collapsed ones included. So one mount re-opened **every** collapsed terminal at once, and `hiddenThreadKeys` was empty within a tick of any mount — meaning the live dot and "Show running integrated terminal" label could never persist either. Attaching no longer touches `panelHidden`.

**Auto-started env actions were killed on quit with no prompt at all.** Only the Run-button path passed an owner, so an action started by *applying an environment* had no descriptor, was skipped by `listRunningDetachedCommands`, and never reached the quit blockers — while `stopAllCodexEnvironmentDetachedCommands` hard-killed it anyway. If it was the only work running, `count` was 0 and quit took the "no active work" fast path. The launchpad spawns its action *before* the thread exists, so the owner is now registered without a thread id and back-filled by `attachDetachedCommandThreadId` once `startThread` returns.

**The heap-snapshot toast invited users to leak their API key.** `safeStorage` decrypts the xAI key and every bot token into main's heap, and this PR added a one-click *main-process* snapshot — then told the operator to "Copy this notice to hand off the snapshot paths," documented as "pasted straight into a bug report or an agent session." Every surface now leads with what the file contains, and the capture is gated on Developer Mode. (File modes were already `0600` — I measured. The exposure was social, not filesystem.)

### Also fixed

- **Close-during-spawn was silently dropped** — the shell survived being dismissed and the renderer re-adopted it from the broadcast. Pending closes are honored on spawn, scoped to an in-flight spawn so they can't shoot down a later terminal.
- **Pane `cwd` flip tore down the xterm** (`undefined` → `homedir()` for a directory-less thread), discarding keystrokes typed during the spawn. `cwd` is create-time only; it lives in a ref now.
- **Reveal-after-exit spawned a phantom shell in `~`** — a fresh quit blocker, conjured by trying to look at one. Reveal is refused when there's no session.
- **IPC inside a `setState` updater** (StrictMode double-fires it).
- **One `destroyed` listener per session** put 10+ on a single WebContents — exactly this feature's load — tripping Node's max-listeners warning.
- Countdown copy said "above"; buttons are below. Chip now matches the pin chip's `role="img"` + `useViewportTooltip`.

### Countdown

Scales with row count (10 terminals is not a ten-second read), and any **deliberate** interaction cancels it. Notably *not* `focusin` or `pointermove`: the Quit button is autofocused, and a pointermove fires when the window merely appears under a stationary cursor. Both cancelled the countdown before anyone had read a word — which would have left an unattended shutdown (OS restart, update install) waiting forever for a human who wasn't there. I only found that because of the new E2E.

The dialog also gets a `default-src 'none'` CSP and denies non-signal navigation.

### Verification

The quit dialog had **no E2E at all** — a `data:` page that signals main by navigating a fake scheme fails silently and still looks perfect. [`quit-confirmation-dialog.spec.ts`](apps/desktop/e2e/quit-confirmation-dialog.spec.ts) now drives the real thing: asserts the countdown is *running* on open, that a keystroke cancels it, that blockers render as links, and that "Stay Open" actually reaches main.

Plus [`keeps a collapsed terminal collapsed across a thread view unmount`](apps/desktop/e2e/integrated-terminal.spec.ts) (fails on the previous commit), and a `counts an auto-started environment action as a quit blocker` unit test.

4629 unit tests; desktop E2E 163 passed (the one failure, `skill-autocomplete-interactions:53`, is the known macOS-only caret flake that fails on `main` too). typecheck, ESLint (0 errors), dependency-cruiser, colors, SQL, codex-storage all clean.

PR #1091, opened 2026-07-31
fix(desktop): isolate thread title helper context

huntharo:
## Summary

- run each Codex title job in a fresh ephemeral thread while pooling only the App Server process/connection
- apply a modern ring-zero helper profile: empty base instructions, no environments or dynamic tools, and explicit suppression of apps, plugins, skills, hooks, orchestrator MCP, web search, reminders, fan-out, memories, goals, multi-agent, image generation, and related harness features
- disable every effective configured MCP server by name and reject the helper before `turn/start` if any thread-scoped MCP server exposes tools, including servers absent from the earlier configuration snapshot
- redact the internal `config/read` response before protocol observers see it, retaining server names only so commands, env, credentials, and secrets cannot enter raw captures or protocol logs
- unsubscribe every helper thread and never use deprecated rollback
- preserve normal interactive and multi-turn Codex behavior outside the one-shot helper path

## Root cause

The observed 7,009- and 9,177-token title jobs were each the first and only turn on distinct threads, so this was not rollback/history accumulation. The title prompts themselves were only 1,817 and 1,527 characters. Most input was cold Codex harness context: plugin/recommended-plugin guidance, capability guidance, and MCP/tool metadata that the old partial suppression profile did not remove. The larger second envelope followed the model change and was not prior helper history.

PwrSnap PR #322 supplied the process-pooling/fresh-thread pattern. A comparison with current OpenClaw found its newer ring-zero profile and post-start MCP checks, which are adapted here to PwrAgent conventions.

## Global AGENTS.md limitation

Current Codex has no thread-level switch for the process-owned global `CODEX_HOME/AGENTS.md`. OpenClaw bypasses it only for private stdio jobs by launching another process with an empty temporary `CODEX_HOME` and separately bridging authentication. A plain isolated home is unauthenticated in the PwrAgent default profile, and PwrAgent repository policy correctly forbids reading or copying Codex-owned auth storage.

The helper therefore retains one bounded global instruction source, reports its count without logging its path or contents, and still runs the title turn. Project instructions remain excluded through the isolated helper cwd and `project_doc_max_bytes: 0`. Fresh threads prevent the global block from accumulating across jobs.

An earlier review-follow-up revision failed closed on any `instructionSources` entry. Real thread `019fba38-7791-7fd1-9299-f47825eb58e6` proved that behavior was a product regression: no helper turn or usage row was created and PwrAgent silently used the truncated prompt fallback. Commit `4702dea3` restores the bounded model turn.

## Security follow-up

- MCP credentials remain protected when `PWRAGENT_PROTOCOL_CAPTURE=1` or `PWRAGENT_APP_SERVER_PROTOCOL_LOG` is enabled: the observer wrapper reconstructs the matching `config/read` response as names-only before forwarding either the parsed envelope or raw JSON.
- The post-start attestation rejects every tool-bearing MCP server, not only names from the pre-start snapshot. This closes the race and managed/virtual-server gap identified in review.
- Current Codex lists disabled configured MCP names in status results, but their tool maps are empty because the connection manager filters `enabled:false` servers. The attestation checks the model-facing capability rather than server-name presence.

## Validation

- `pnpm test apps/desktop/src/main/__tests__/codex-client.test.ts` — 121 passed
- `pnpm typecheck`
- targeted ESLint — 0 errors (3 existing warnings)
- `pnpm lint:boundaries`
- `pnpm lint:codex-storage`
- `pnpm --filter @pwragent/desktop build`
- `git diff --check`
- zero-token App Server probes for account and MCP isolation behavior
- minimal live App Server title turn: generated `Three Bad Brothers’ Origin Story`, 1 bounded global instruction source, 0 tool-bearing MCP servers, and successful unsubscribe

### DIFF
--- a/apps/desktop/src/main/__tests__/backend-registry.test.ts
+++ b/apps/desktop/src/main/__tests__/backend-registry.test.ts
@@ -13409,7 +13409,7 @@ command = "pnpm dev"
     const titleService = {
       generateTitle: vi.fn(async () => ({
         status: "invalid" as const,
-        reason: "title_too_many_words",
+        reason: "title_empty",
         helperThreadId: "invalid-title-helper-thread",
         helperTurnId: "invalid-title-helper-turn",
         model: "gpt-5.6-luna",
@@ -13468,7 +13468,7 @@ command = "pnpm dev"
         monitorTurnId: "invalid-title-helper-turn",
         preferredModel: "gpt-5.6-luna",
         preferredReasoningEffort: "low",
-        lastMessage: "Title generation failed: title_too_many_words",
+        lastMessage: "Title generation failed: title_empty",
         monitorUsage: expect.objectContaining({
           model: "gpt-5.6-luna",
           tokenUsage: {
--- a/apps/desktop/src/main/__tests__/codex-client.test.ts
+++ b/apps/desktop/src/main/__tests__/codex-client.test.ts
@@ -135,6 +135,11 @@ class MockTransport implements JsonRpcTransport {
     },
   };
   static configReadError: { code?: number; message: string } | undefined;
+  static threadMcpServerStatusResult: unknown = {
+    data: [],
+    nextCursor: null,
+  };
+  static mcpServerStatusError: { code?: number; message: string } | undefined;
   static lastConfigValueWritePayload: unknown;
   static rateLimitsResult: unknown = {
     rateLimitsByLimitId: {}
@@ -169,6 +174,7 @@ class MockTransport implements JsonRpcTransport {
     const payload = JSON.parse(message) as {
       id?: string;
       method?: string;
+      params?: Record<string, unknown>;
     };
 
     if (payload.method === "initialize") {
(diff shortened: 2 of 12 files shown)


huntharo:
Follow-up from the live dev-profile reproduction:

- The helper run completed in about 2.4 seconds with 1,879 uncached input tokens and 24 output tokens; it did not hang.
- It was rejected as `title_too_many_words`, even though the prompt described six words as a preference. The helper now enforces the real 50-character ceiling in its output schema but accepts a clear seven-word title.
- The Pricing panel terminal-status set omitted the actual `failed` status. That made a finalized failed helper render as Live indefinitely while the Sub-agents panel correctly showed Failed. Pricing now treats `failed` and `blocked` as terminal, with coverage for every terminal status.

Added in `912f1598`. Focused suites pass (598 tests before the final focused refinement; 74 tests after it), plus typecheck, targeted ESLint, dependency boundaries, Codex storage lint, desktop build, and diff checks.

huntharo:
Title-length policy follow-up in `cdc5dce4`:

- The helper prompt still targets 6 words and 50 characters.
- PwrAgent accepts a strict 30% overage: 7 words and 65 characters.
- Responses beyond either tolerance are locally truncated to those limits and treated as generated titles, not failed title-helper runs.
- The structured-output schema no longer rejects over-limit strings before PwrAgent can recover them.
- Prompt/cache version advanced to `thread-title-v2`.

Validation: 477 focused tests passed, follow-up title tests passed, typecheck passed, targeted ESLint has 0 errors (1 unrelated existing warning), and the desktop production build passed.

huntharo:
Pricing-label follow-up in `4aea2820`:

Known monitor-scope rows now use their durable sub-agent identity instead of all reading `Sub-agent usage`:

- system title helper → `Thread naming`
- PwrAgent task monitor → `Monitor usage`
- code review → `Review usage`
- Codex native spawn → `Codex sub-agent usage`

Legacy/orphaned rows without a matching sub-agent summary retain the generic fallback. Renderer test (57 tests), targeted ESLint, desktop typecheck, and diff checks pass.

PR #1146, opened 2026-08-02
test(desktop): add macOS VM E2E coverage

huntharo:
## Summary

- Add the Tart macOS VM lab skill and contributor guidance so headed Electron E2E runs stay off the host desktop.
- Add a guarded macOS/ARM64 self-hosted CI lane targeting the selected-repository PwrDrvr macOS runner group.
- Add Git LFS-managed macOS visual goldens and two replay-backed visual-regression tests.
- Transport HEAD-referenced LFS objects directly for private `--local` VM runs, without requiring the guest Git remote to act as an LFS server.
- Disable Electron GPU use only for the Tart E2E lane to avoid AppleParavirtGPU instability while preserving ordinary host GPU coverage.
- Stabilize three pre-existing E2E interactions exposed by macOS timing and keyboard semantics.
- Give visual E2E a fixed `1.2.3-beta.1` application version so direct Electron launches cannot leak the Electron runtime version into screenshots.
- Align the dependency, native rebuild, and packaged Electron runtime on 41.10.3, with a CI policy that rejects future lockfile/electron-builder drift.

## Why

PwrAgent needs reliable macOS desktop coverage and deterministic visual checks without stealing an operator’s desktop focus. The shared runner group is restricted to PwrAgent and PwrSnap, and fork-head PRs remain excluded from the self-hosted job.

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm lint:electron-version`
- `pnpm test scripts/check-electron-version-policy.test.mjs apps/desktop/src/main/__tests__/app-version.test.ts` — 4 passed
- `pnpm typecheck`
- `pnpm lint:boundaries`
- Targeted ESLint: no errors
- Rebased macOS VM visual spec: 2 passed, 0 failed, 0 skipped; existing goldens unchanged
- Final combined Tart regression run: 24 passed, 0 failed, 0 skipped
- `bash -n` across the VM-lab scripts
- `git diff --check`
- `git lfs fsck`


huntharo:
Measured this branch in the Tart dev VM while chasing the vmapple VideoToolbox kernel leak for #1168, and there's one result you'll want before flipping the lane on.

**The `PWRAGENT_E2E_DISABLE_GPU` gate does not stop the kernel leak.** `disableHardwareAcceleration()` + `--disable-gpu` switch rendering to SwiftShader, but they don't stop the GPU process from initializing hardware *media codecs* — and every VideoToolbox init inside a Virtualization.framework guest leaks an `AppleVideoToolboxParavirtualizationUserClient` that the vmapple driver never frees, even at process death.

Full suite (199 tests, `workers: 1`), guest rebooted before each run so the kernel table starts clean, `ioclasscount AppleVideoToolboxParavirtualizationUserClient` before → after:

| branch | `PWRAGENT_E2E_DISABLE_GPU` | before → after | delta | specs |
|---|---|---|---|---|
| `main` (pre-#1168) | unset | 2 → 176 | +174 | 165 passed, 3 failed |
| **this branch alone** | **`1`** | **2 → 178** | **+176** | **170 passed, 0 failed** |
| this branch + #1168 | `1` | 2 → 5 | **+3** | 170 passed, 0 failed |

+176 *with* the gate versus +174 with no mitigation at all — the gate makes no difference to this particular failure mode. Past ~1.1k live clients the driver's `IOService::newUserClient` wedges, new Electron helpers hang at birth in-kernel, and app exit blocks in `wait4` on an unkillable helper: the uniform ~+6s/spec teardown that took PwrSnap's jobs from ~3–4 min to ~17 min until the guest was rebooted. At +176/run a guest gets there in ~6 suite runs; at +3/run, ~370. PwrSnap already contributes ~143/job to the same guests, so enabling this lane unmitigated would roughly halve runs-per-boot for both repos.

**This is already fixed on `main`** — #1168 merged as 9981e9f4c and adds the two codec switches at the fixture's Electron launch site. This branch predates that merge, so **merging or rebasing on `main` picks it up**, and the third row above is exactly that combination (I cherry-picked it onto this branch to measure). No conflicts.

Everything else here looks ready from the VM's perspective:

- **Suite is green**: 170 passed, 0 failed, 31 skipped. Your spec fixes close the three that fail on pre-#1168 `main` in a VM (`directory-launchpad-workspace:757`, `live-work-rail-sticky-gap:14`, `skill-autocomplete-interactions:53`) — I confirmed those three fail on `main` and pass here.
- **Teardown telemetry at baseline**: 3 `graceful close failed` warnings out of 170 in both runs, matching PwrSnap's healthy-guest baseline of 0–3. The bounded-close + process-tree-kill path does its job.
- **No duration cost**: 4.7m for this branch alone, 4.8m with the codec switches.

Background and the original PwrSnap investigation (including the false-start diagnosis worth not repeating): `docs/solutions/2026-08-03-e2e-teardown-login-shell-hang.md` in PwrSnap, fix in pwrdrvr/PwrSnap#366.

One lesson from that write-up that applies directly here: a fix for this class of degradation has to be validated against a *degraded* guest or across the full onset window. A fresh guest is always fast, so a clean run on a freshly booted VM proves very little on its own.

PR #735, opened 2026-06-10
feat(federation): complete remote instance federation

huntharo:
## Summary

Completes the U1-U9 instance-federation workflow: gateway and client instances can enroll over the authenticated WebSocket transport, discover every connected federation member, open branded remote windows, operate remote threads with live events, search across instances, and route messaging bindings to remote threads through the gateway relay.

## What changed

- Completes the federation backend surface for navigation, thread creation/fork/review, turn start/steer/interrupt/compact, pending requests and approvals, model/runtime/execution settings, skills, environment selection/actions with setup progress, and workspace handoff.
- Relays client-to-client RPC, backend events, and environment progress through the gateway with capability checks, deadlines, loop prevention, reconnect backoff, duplicate-session replacement, and active revocation.
- Adds federation-aware thread search with source identity, remote-window selection, partial-failure reporting, and no local fallback.
- Routes messaging browse/resume/bind/status/turn controls/settings/approvals to remote owners while keeping callback payloads provider-neutral and remote thread identities collision-safe.
- Expands Settings with Cloudflare Access service-token and mTLS credentials, peer health, unavailable reasons, audit diagnostics, revocation, and connected-only remote-window actions.
- Brands remote windows and thread chrome with the authoritative instance label.
- Adds `docs/federation.md` with local dogfood, Cloudflare Tunnel, Access, mTLS, credential, diagnostics, revocation, and threat-model guidance.

## Test locally

Build the desktop app, then launch isolated gateway and client profiles:

```bash
pnpm --filter @pwragent/desktop build
PWRAGENT_PROFILE=gateway pnpm --filter @pwragent/desktop preview
PWRAGENT_PROFILE=client pnpm --filter @pwragent/desktop preview
```

Use Settings -> Federation on the gateway to generate an invite, import it on the client, then use the connected peer list to open either instance. Detailed direct-local and Cloudflare setup is in `docs/federation.md`.

Cloudflare provisioning remains an operator-managed deployment step; PwrAgent stores and applies the configured Access/mTLS credentials but does not create tunnels or Access policies.

## Validation

- `pnpm test` (388 files, 4,874 tests)
- `pnpm typecheck`
- `pnpm lint:eslint` (0 errors; repository warning baseline remains)
- `pnpm lint:boundaries`
- `pnpm lint:codex-storage`
- `pnpm lint:colors`
- `pnpm lint:sql`
- `pnpm licenses:check`

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
GPT-5

PR #1202, opened 2026-08-05
fix(desktop): federation UX cleanup after dogfood review

huntharo:
## Summary

Federation merged (#735, #1191) but wasn't done. This PR addresses the first-day dogfood review findings — the four reported issues plus the gaps a full code sweep surfaced.

### Reported issues

1. **Activity feed grows forever** — `federation_session_audit` had no retention (a broken client wrote ~5,700 rows/day) and the "50" chip silently saturated at the fetch limit.
   - `appendAudit` now collapses identical consecutive events into one repeat-counted row (`×N`, "Repeated N times since …"), shown in the Diagnostics list.
   - The table gets a 500-row FIFO cap in `StateDb.cleanupExpired`, mirroring `messaging_activity_log`.
   - The chip reads `50+` when the view is at its fetch limit.
2. **No entry point outside Settings** — added **File → Remote Instances** listing connected peers (opens the remote-threads window; hidden when no peers, kept fresh on peer status changes). The ⌘⇧F results header now discloses remote scope: `Remote: Studio Mac (3) — title matches only`, including per-peer failures. (⌘K quick-jump stays intentionally local — it filters the current window's in-memory list.)
3. **Client mode leaves gateway-only fields enabled** — Listen host/port/Public URL disable outside gateway/dual; Gateway URL disables outside client/dual, with explanatory sub-copy. Disabled, not hidden, so `dual` stays coherent.
4. **Pairing is invisible and stuck** —
   - New **Gateway Enrollment** section: pinned gateway (friendly name), gateway URL, invite-import time, pending-invite state — an enrolled-but-failing client no longer looks identical to a never-enrolled one.
   - **Forget gateway** (two-step confirm) clears the pinned identity/signing/Noise keys and pending token via new `federation:reset-enrollment` IPC; a client-only instance drops back to disabled mode instead of a doomed retry loop.
   - Auth-class failures (unknown/revoked peer, invite reuse, pin mismatch, **protocol/transport version skew** — the exact failure in the screenshots) now report status **Rejected** with per-cause remediation copy instead of "Connecting" forever.
   - Import success now shows feedback instead of rendering identically to pre-import.

### Additional findings fixed

- **Remote windows leaked local chrome**: Settings gear, Automations button, and the MSG messaging-status chip are hidden in federation windows (they showed *local* state in a window branded as another machine), and main-process Settings/New Thread/Replay Onboarding menu pushes skip federation windows.
- **GUID soup → names**: new `[federation] instance_label` config (Settings → "Instance name", defaults to machine hostname) advertised on connect and refreshed on reconnect; peer cards lead with the name, GUID demoted to metadata; diagnostics rows resolve peer ids to names.
- **`messaging_route` was declared but dead**: now advertised in `DEFAULT_CAPABILITIES` and enforced at the messaging bridge seam (chat surfaces cannot browse/drive a peer that doesn't grant it) — the fail-closed gate the plan's KTD3 called for. The orphaned `federated-messaging-routing.ts` module is removed.
- **`remote_window` now actually gates windows**: Browse button, open-window IPC, and the new menu all check it.
- **Federated search hardening**: 10s per-peer timeout (was: one hung peer stalls ⌘⇧F for the full 30s RPC deadline) and `searchedInstances` reporting.
- **Remote RPC errors** no longer drop the remote's human-readable message (previously surfaced as bare codes like `unknown_peer`).
- **Settings polish**: invite generation auto-copies and renders as a code chip + Copy button (same pattern as messaging pairing) instead of a select-it-yourself textarea; `.settings-button-row` gets vertical padding so buttons stop touching section borders.

### Tests

- New: audit repeat-collapse + GC cap, failure classification, search timeout + instance summaries, Remote Instances menu, mode-aware field gating, enrollment display + forget flow, `instance_label` TOML round-trip.
- Full workspace suite: **6012 passed**; `lint:eslint` (0 errors), `lint:sql`, `lint:colors`, `lint:codex-storage`, `lint:boundaries`, `typecheck` all green.

### Notes

- "Unexpected federation auth challenge" from the review screenshots no longer exists in the tree — it was a pre-#1191 client hitting a post-#1191 Noise gateway. The new version-skew remediation copy covers that scenario explicitly.
- Remote messaging state (a real remote MSG surface) and Noise-over-SSH / multi-path peer endpoints are deliberate follow-ups, not in this PR.


huntharo:
Second dogfood round (00a17c4d5), from live testing findings:

- **Silent federation key rotation fixed** — root cause of "the key was invalid after having previously worked". Undecryptable stored keys (dev-build keychain identity changes) now fail loudly with remediation instead of silently minting new keys that break every pinned pairing. Forget-gateway on a pure client clears local key material so re-pairing mints fresh keys.
- **Generate invite disabled in client mode** (UI + runtime guard); importing an invite on gateway/dual now preserves the listener (→ dual) instead of demoting to client.
- **Save federation settings surfaces write failures** instead of silently keeping unsaved state.
- **Integrated terminal hidden for remote threads** — it opened a local shell in a peer-branded window.
- **Pin/unpin routed over federation** (`backend.setThreadPin`, `thread_navigation` cap) — remote unpin previously wrote a phantom row into the viewer's overlay store and reverted.
- **PR polling stays on the owning instance** — remote windows no longer run local `gh` lookups for remote threads (which used the viewer's credentials against remote thread ids), no longer trigger a never-converging 60s remote-snapshot refresh loop, and no longer clobber the local poller's focus set. Audit confirmed the PR watcher itself never double-fires (watch rows + turn submission are owner-local); detach/auto-dispatch controls refuse remote threads pending federation routing.

Full suite: 6012 passed. Lint/typecheck/boundaries green.

huntharo:
Self-review pass (ac03b9431) caught and fixed one **blocker** before merge:

- **Upgrade break**: adding `messaging_route` to `DEFAULT_CAPABILITIES` made upgraded clients request a capability absent from every pre-existing enrollment row, and the reconnect policy's all-or-nothing check rejected the whole session with `capability_denied` — existing pairings would have hard-failed after both sides upgraded. Reconnect now grants the requested∩stored intersection (never more than enrollment authorized; zero overlap still rejects). Regression test added for the exact upgrade scenario.
- Also guarded pin **drag-reorder** in remote windows (`reorderThreadPins` isn't federation-routed yet — same phantom-write class as the unpin bug). Proper routing rides the follow-up card with detach/auto-dispatch.

Remaining review notes were nits (MessagingStatusBar effects still fire once behind its null render in remote windows; a duplicated helper call in Connection JSX; per-op peer lookup in the messaging bridge) — left as-is deliberately.

Full suite: 6013 passed.

huntharo:
Capability model redesigned per operator direction (37984b686) — pre-release, so semantics locked in now:

- **Enrollment is identity trust, not a capability allowlist.** An enrolled peer gets whatever its build advertises; the stored row's capabilities are informational and refresh on reconnect. Per-peer narrowing is deferred to the future RBAC work. This replaces the earlier intersection-negotiation fix.
- **Forward-compatible negotiation locked in at the transport layer**: capability lists validate structurally, signatures cover the raw list as transmitted, and unknown names from newer builds are filtered *after* verification instead of failing the handshake. Adding capability #11 post-release is now a non-event for mixed-version pairs (covered by a live two-instance test advertising `capability_from_the_future`).
- **`reorderThreadPins` routed over federation** (failing test first) — remote drag-reorder now lands on the owning instance instead of being suppressed.
- Nits fixed: MessagingStatusBar fully inert in remote windows, remediation helper computed once, federation-window PR-refresh defense throws instead of returning a diffable empty result.

Full suite: 6014 passed; typecheck/eslint/boundaries green.

PR #1205, opened 2026-08-05
feat(desktop): copy messages with a rich-text HTML clipboard flavor

huntharo:
## Problem

Copy message wrote only the raw markdown to the clipboard, so pasting into rich-text targets (Google Docs, Gmail, Word) produced literal markdown syntax (`**bold**`, `### headings`, pipe tables). Reported against a Grok ACP thread, but generic to the renderer's copy path — every backend was affected.

## Approach

No Swift helper needed (unlike PwrSnap): Electron's `clipboard.write({ text, html })` writes multiple pasteboard flavors in one call. Verified on macOS that a single write lands both `text/plain` and `text/html` (`clipboard.availableFormats()` → `["text/plain", "text/html"]`), which is exactly the Cmd+V (rich) / Cmd+Shift+V (plain markdown) split.

- **Preload**: new `copyRichText({ text, html })` bridging to `clipboard.write`.
- **`copyTextWithHtml`** (renderer lib): prefers the bridge, falls back to `navigator.clipboard.write` with a two-flavor `ClipboardItem`, then to the existing plain-text path.
- **`markdown-clipboard-html.ts`**: renders the message markdown to standalone semantic HTML via `renderToStaticMarkup` + `react-markdown` with the same source repairs (`repairNestedLanguageFences`, `protectComposerHyphenListItems`) and `remark-breaks`/`remark-gfm` plugins ThreadMarkdown uses. App-specific plugins (PR chips, table profiling, math) are intentionally excluded — their output only makes sense inside the renderer.
- **`TranscriptCopyButton`**: optional lazy `html` prop; HTML is rendered only on click. Only the message-level Copy message button passes it — code-block and path copy buttons stay plain-text only.

## Testing

- New unit tests: `copyTextWithHtml` bridge/fallback ordering, markdown→HTML rendering (emphasis, headings, GFM tables, fenced code, nested lists), and a transcript-list test asserting Copy message sends both flavors.
- Existing copy tests unchanged and passing (plain-text fallback preserved when no rich bridge exists): full desktop renderer suite 109 files / 1852 tests green.
- `pnpm typecheck`, `pnpm lint:eslint`, `pnpm lint:boundaries` clean.
- Electron smoke test on this machine confirmed both flavors on the pasteboard from one write.

### DIFF
--- a/apps/desktop/src/main/__tests__/index.test.ts
+++ b/apps/desktop/src/main/__tests__/index.test.ts
@@ -20,6 +20,8 @@ const registerAutomationIpcHandlersMock = vi.fn();
 const disposeAutomationIpcHandlersMock = vi.fn();
 const registerAppMetadataIpcHandlersMock = vi.fn();
 const disposeAppMetadataIpcHandlersMock = vi.fn();
+const registerClipboardIpcHandlersMock = vi.fn();
+const disposeClipboardIpcHandlersMock = vi.fn();
 const registerAppUpdateIpcHandlersMock = vi.fn();
 const disposeAppUpdateIpcHandlersMock = vi.fn();
 const initAutoUpdaterMock = vi.fn();
@@ -250,6 +252,11 @@ vi.mock("../ipc/automation-ipc", () => ({
 vi.mock("../ipc/app-metadata", () => ({
   registerAppMetadataIpcHandlers: registerAppMetadataIpcHandlersMock,
   disposeAppMetadataIpcHandlers: disposeAppMetadataIpcHandlersMock,
+}));
+
+vi.mock("../ipc/clipboard", () => ({
+  registerClipboardIpcHandlers: registerClipboardIpcHandlersMock,
+  disposeClipboardIpcHandlers: disposeClipboardIpcHandlersMock,
 }));
 
 vi.mock("../auto-updater", () => ({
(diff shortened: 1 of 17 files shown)


huntharo:
**Follow-up commit 66729e31c — the original preload approach did not work in the real app.**

Field report: pasting into Google Docs still produced plain markdown from a dev build that included the fix. Root cause found by reproducing the app's exact window config (`sandbox: true`, `contextIsolation: true`):

- In a sandboxed preload, `require("electron")` exposes only `contextBridge`, `crashReporter`, `ipcRenderer`, `nativeImage`, `sharedTexture`, `webFrame`, `webUtils` — **no `clipboard`**. Both `copyRichText` and the pre-existing `copyText` bridge threw on every call; plain copy only ever worked because the renderer silently fell back to `navigator.clipboard.writeText`.
- The rich `navigator.clipboard.write` fallback also failed, so the copy degraded to plain text.

Fix: clipboard writes now go over IPC (`clipboard:write-text`, `clipboard:write-rich-text`) to main-process handlers in `src/main/ipc/clipboard.ts`, matching how every other privileged preload call works. This also repairs the long-silently-broken `copyText` bridge. Verified under a `sandbox: true` window that a single IPC round-trip lands both `text/plain` and `text/html` on the macOS pasteboard.

PR #1217, opened 2026-08-05
fix(desktop): give federation windows a proper remote-identity marker

huntharo:
## Summary

Dogfood follow-ups on remote-window presentation (screenshots in the report):

- **No more raw `pwr_<uuid>` chip** in the thread title bar — the thread-header federation chip rendered `instanceLabel`, which falls back to the raw instance id when the peer never set a label. Removed; the window itself is the remote branding.
- **Profile-style identity pill in the left bar** — replaces the truncated `Remote · <label> · <tag>` masthead suffix. It sits in the slot the local runtime/profile rows vacate in remote windows, shows the full machine name, carries a viewport tooltip with the instance id (the old `title` attr never expanded usefully), and click-copies the instance id.
- **Title-bar badge when the sidebar is hidden** — the pill leaves with the sidebar, so the relocated masthead (macOS/Linux) and the Windows `AppTitleBar` render a compact `FederationRemoteBadge` next to the brand. The window always has a visible remote marker.
- **Masthead leak fix**: the relocated `MastheadActions` copy (shown when the sidebar is hidden) still rendered Open settings / Open automations in remote windows — collapsing the sidebar resurfaced local Settings inside a peer-branded window. Now gated like the sidebar copies from #1202.

## Verification

- Renderer suite: 1,885 tests pass; the sidebar remote-window test now asserts the pill (full name, no instance id in the visible text, copy affordance).
- Federation E2E extended: hides the sidebar in the remote window and asserts Settings/Automations stay hidden, the Search button remains, and the remote badge appears; passes locally against a built `out/`.
- Also tightened the spec's `Connected` locator to exact match — #1203's endpoint rows added an `Active · Connected …` line that made the substring match strict-mode ambiguous (latent flake).
- Typecheck, ESLint (0 errors), `lint:boundaries` pass.

## Notes

- The client-side label for an enrolled gateway defaults to "Gateway"; operators can set a friendlier name via Settings → Federation → Instance name on the peer.
- The `.sidebar__federation-label` class stays on the pill's label span as the stable E2E marker (documented in app.css).

### DIFF
--- a/apps/desktop/e2e/federation-remote-window.spec.ts
+++ b/apps/desktop/e2e/federation-remote-window.spec.ts
@@ -186,11 +186,13 @@ test.describe("federation remote window", () => {
       // menu.
       await expect
         .poll(async () =>
-          await electronApp.evaluate(({ BrowserWindow }) =>
-            BrowserWindow.getAllWindows().map((win) => win.getTitle()),
-          ),
+          (
+            await electronApp.evaluate(({ BrowserWindow }) =>
+              BrowserWindow.getAllWindows().map((win) => win.getTitle()),
+            )
+          ).some((title) => /^PwrAgent - ./.test(title)),
         )
-        .toContain("PwrAgent - Gateway");
+        .toBe(true);
 
       // The peer's threads render; the local thread does not leak in.
       await expect(
--- a/apps/desktop/e2e/federation-remote-window.spec.ts
+++ b/apps/desktop/e2e/federation-remote-window.spec.ts
@@ -180,6 +180,18 @@ test.describe("federation remote window", () => {
       expect(target?.scope).toBe("remote");
       expect(target?.instanceId).toBe(gateway.instanceId);
 
+      // The OS-level window title must keep the peer label — the
+      // renderer's static <title> used to clobber it back to the app
+      // name, collapsing every window to one entry in the macOS Window
+      // menu.
+      await expect
+        .poll(async () =>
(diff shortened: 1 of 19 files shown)


huntharo:
Added `3a3d6eab0`: remote window titles now keep the "PwrAgent - <machine>" label in the macOS Window menu. The renderer's static `<title>` (index.html) was replacing the BrowserWindow title on load, collapsing every window to the same menu entry — federation windows now preventDefault the page-title swap at the BrowserWindow level, and the federation E2E asserts the OS-level title.

huntharo:
Two more commits:

- `0f158c4a6` — review-nit polish: sidebar pill ellipsis specificity, copied-timer cleanup on unmount, Windows shows the title-bar badge only while the sidebar is hidden, E2E title assertion decoupled from the default peer label, and a FederationRemoteBadge unit test (copy + tooltip).
- **Profile-aware peer labels**: peers now display as `<machine> / <profile>` when the profile isn't `default`, or whenever several profiles of the same machine are connected (then even `default` shows). A lone default-profile peer keeps the bare machine name. Clients now actually advertise their profile on connect — the transport carried `profileName` end-to-end but `connectClient` never passed it. Composed labels flow through window titles, the sidebar pill/badge, the Remote Instances menu, Settings peer rows, thread rows, and federated-search chips from one shared helper (`formatFederationPeerDisplayLabel`, unit-tested).

PR #1226, opened 2026-08-05
fix(desktop): graceful disconnected state for remote windows + observable federation evictions

huntharo:
## Summary

Fixes the dogfood report: a remote window whose peer went down spewed `scheduled-actions:list … peer not connected` every 5s over a blank thread list with a live composer, and after the peer came back the window flashed connected/disconnected on an exact 30s cycle indefinitely.

**Root cause of the 30s flap** (found by code investigation): the M5 was running two profiles whose `state.db` shares a lineage — the federation **instance id and identity key both live in `state.db`**, so a cloned profile presents the *same identity*. The gateway evicts the previous connection whenever the same peer id authenticates (`federation-transport.ts` replace path), so the two processes evict each other forever; because no session survives the 60s stability window, reconnect backoff pins at the 30s ceiling (`min(1000·2^n, 30000)` never resets). The eviction was completely unobservable: bare `socket.close()` (code 1000, no reason), zero logging, and the client's close handler dropped the code/reason and logged nothing.

### Renderer — graceful disable (option 1 from the report)
- `useFederationPeerConnectivity`: seeds from federation health, follows `federation/peerStatus/changed` (the signal `useThreadNavigation` already uses).
- Scheduled-actions projection **suspends its 5s poll** while the peer is down (resume refreshes immediately) and warns **once per distinct failure** — kills the log spam.
- Remote windows go explicitly read-only on disconnect: **composer disabled** + a warning banner naming the machine ("… is unreachable — reconnecting. Threads shown may be stale."). The runtime keeps reconnecting on its own.

### Main — observable evictions + correctness
- Gateway eviction now audits `replaced` for the old session *before* the new `connected` row (causal order), logs with replacement counts, closes with application code **4001 `replaced_by_new_session`**, and appends `duplicate_instance_id_suspected` when evictions repeat ≥3× in 5 minutes. No more duplicate `transport_closed` audit row for the evicted socket.
- Client close path now carries the WebSocket code/reason: logs `federation client session closed` with code/reason/**session age** (a repeating sub-minute age makes a kick loop obvious), audits `transport_closed:<code>`, and maps 4001 to a new `replaced` failure kind — health reads **degraded** with an actionable message: duplicate identity from a cloned profile; reset federation on one of them.
- `disconnectAdvertisedPeers` no longer false-publishes "disconnected" for peers holding a live connection to this instance's own gateway (dual-mode bug).

## Verification

- Transport unit test: same-id reconnect evicts with 4001 + correct audit ordering, no duplicate rows.
- Connectivity hook unit tests (3): local windows always connected, seed + transitions, other-peer isolation.
- Federation E2E extended: **kills the in-process gateway** → banner + disabled composer appear; **restarts it on the same port/identity** → window heals with no reload. Passes locally.
- Workspace typecheck, ESLint (0 errors), `lint:boundaries`, renderer lib + main federation suites green.

## Notes / follow-ups (not in this PR)
- No heartbeat/keepalive exists — an abruptly dead peer stays "connected" until TCP gives up (RPCs die at the 30s deadline). `expireStaleSessions` exists but has no caller; wiring it to a ping/pong is the natural next PR.
- Enrollment currently re-pins an existing peer id to a new key without objection; rejecting that unless the operator re-pairs would close the H2 variant of the duplicate-identity case.
- Operator remediation for the observed loop: one of the two same-machine profiles has a cloned `state.db` — reset federation on one of them (Forget gateway + re-enroll); it will mint a fresh identity.

### DIFF
--- a/apps/desktop/e2e/federation-remote-window.spec.ts
+++ b/apps/desktop/e2e/federation-remote-window.spec.ts
@@ -100,7 +100,7 @@ async function createLocalControlFixture(): Promise<{
 
 test.describe("federation remote window", () => {
   test("enrolls, browses remote threads, pins on the owner, and hides local-only chrome", async () => {
-    test.setTimeout(180_000);
+    test.setTimeout(300_000);
 
     let gateway: InProcessFederationGateway | undefined;
     const fixture = await createLocalControlFixture();
@@ -347,6 +347,31 @@ test.describe("federation remote window", () => {
       expect(
         gateway.calls.map((call) => call.method),
       ).not.toContain("refreshThreadPullRequests");
+
+      // Peer death: the window flips to an explicit read-only state —
+      // disconnected banner, composer disabled — instead of a half-dead
+      // surface hammering the peer with failing RPCs.
+      await gateway.stop();
+      await expect(
+        remote.locator(".federation-disconnected-banner"),
+      ).toBeVisible({ timeout: 30_000 });
+      await expect(
+        remote.locator(".composer-tiptap-input__editor"),
+      ).toHaveAttribute("contenteditable", "false", { timeout: 15_000 });
+
+      // Recovery: the same identity returns on the same port; the client
+      // reconnects on its own backoff and the window heals without a
+      // reload — banner gone, remote threads back.
+      await gateway.restart();
+      await gateway.waitForNextConnection(90_000);
+      await expect(
+        remote.locator(".federation-disconnected-banner"),
+      ).toHaveCount(0, { timeout: 60_000 });
+      await expect(
+        remote.locator(".thread-row__title", {
+          hasText: "Remote gateway thread one",
+        }),
+      ).toBeVisible({ timeout: 30_000 });
     } finally {
       await app?.close();
       await gateway?.close();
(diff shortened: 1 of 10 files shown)


huntharo:
Restacked onto #1225 (`feat/federation-transport-keepalive`) and retargeted the base — this is now a stacked PR; merge #1225 first and GitHub will retarget this back to `main` when its branch deletes.

Conflict resolution was a clean union in both files: the transport keeps #1225's keepalive/frame-ceiling/auth-deadline machinery alongside this PR's eviction close-codes (the widened `close(code, reason)` signature composes with #1225's `terminate()` for the stale sweep), and the E2E gateway harness keeps #1221's remote-PTY wiring inside this PR's `buildServer`/`restart` structure.

Also includes the review-polish commit: close-handler declaration order, revocation closes with 4002 (client maps it to an auth-kind 're-pair needed' message), and a new `onPeerReplaced` transport callback so duplicate-identity churn surfaces as an `unavailableReason` on the peer — visible in Settings rows and to remote viewers via the relayed peer directory, not just to the evicted process.

Verified on the stack: typecheck, ESLint, transport (incl. #1225's 18 keepalive tests) + runtime + PTY + connectivity-hook suites (57 tests), and the full federation E2E — enroll → browse → remote PTY → gateway kill (banner + disabled composer) → same-identity restart → no-reload recovery. Nice compounding effect: the disconnect phase now also exercises keepalive-era close paths end-to-end. Re-review requested per the plan.

PR #1248, opened 2026-08-05
feat(desktop): federation agent-tool layer

huntharo:
## Summary

Gives agents (in-thread agents, and the Star Map `[+]` intake sub-agent) eyes and hands across federated instances. Brainstorm + plan docs are the first commit; the decision record for authorization is in the plan.

- **New `federation` agent-tool catalog** in the unified `pwragent` namespace, advertised through both Codex dynamic tools and the loopback MCP server:
  - `list_federation_instances` — fleet directory (label, purpose notes, celestial icon, status, capabilities, `isLocal`); returns just the local instance when federation is disabled.
  - `list_instance_projects` — an instance's directories with launchpad availability, via the federation-routed navigation snapshot.
  - `create_instance_thread` — thread creation on any instance through `materializeDirectoryLaunchpad`, inheriting the project's launchpad **settings** presets (never its unsent draft prompt/attachments).
  - `search_federation_threads` — fleet-wide search merging local results with the `searchConnectedPeers` fan-out.
- **`[federation] instance_notes`** — operator-written purpose field ("Studio Mac — PwrSnap dev + screen recording"), wired end-to-end like `instance_label` and advertised to peers via handshake + peer-directory gossip. A shared optional `icon` field rides the same wire surfaces for the Star Map branch to populate (assignment/sync stays there — coordinate on `FederationPeerSummary.icon`).
- **Authorization decision (recorded in the plan):** no new capability analog to `messaging_route`. Agent-originated cross-instance control rides the same per-method capability gates as operator control (`thread_navigation`, `turn_control`, `environment_actions`, `federated_search`); enrollment stays the trust boundary per the #1202 trusted-operator model, and future per-peer RBAC narrows agent traffic for free.
- **`~/.pwragent/AGENTS.md` convention** documented in `docs/agent-operator-preferences.md` and referenced from the tool descriptions so intake agents consult the operator's thread-startup preferences.

The handler lives in federation-land and is injected from `main/index.ts` (the federation runtime already imports the registry, so the reverse import would cycle).

## Sequencing

The Star Map session's `[+]` intake unit should consume these tools instead of inventing its own; its map/icons/topology units are independent. Cmd+K unification can adopt `search_federation_threads`.

## Testing

- `pnpm typecheck`, `pnpm lint:eslint` (0 errors), `pnpm lint:boundaries`, `lint:sql`, `lint:codex-storage`, `lint:colors` all pass.
- Full workspace suite: 6330 passed / 3 skipped.
- New coverage: shared contract tests, tool schema/validation/normalization tests (mirroring the thread-orchestration patterns), handler-service tests (instance listing incl. non-federated, remote-vs-local routing, launchpad settings merge, search merge + instance scoping), enrollment/reconnect notes+icon passthrough (present-but-empty clears, absent keeps), and `instance_notes` TOML read/round-trip/delete-on-empty.

### DIFF
--- a/apps/desktop/src/main/__tests__/desktop-config-federation.test.ts
+++ b/apps/desktop/src/main/__tests__/desktop-config-federation.test.ts
@@ -11,6 +11,7 @@ describe("desktop config [federation] section", () => {
       "[federation]",
       'mode = "gateway"',
       'instance_label = "Studio Mac"',
+      'instance_notes = "PwrSnap dev + screen recording"',
       'listen_host = "127.0.0.1"',
       "listen_port = 47830",
       'public_url = "https://pwragent.example.com"',
@@ -25,6 +26,7 @@ describe("desktop config [federation] section", () => {
     expect(config.federation).toEqual({
       mode: "gateway",
       instanceLabel: "Studio Mac",
+      instanceNotes: "PwrSnap dev + screen recording",
       listenHost: "127.0.0.1",
       listenPort: 47830,
       publicUrl: "https://pwragent.example.com",
@@ -39,6 +41,7 @@ describe("desktop config [federation] section", () => {
       federation: {
         mode: "dual",
         instanceLabel: "Studio Mac",
+        instanceNotes: "PwrSnap dev + screen recording",
         listenHost: "0.0.0.0",
         listenPort: 47831,
         publicUrl: "https://gateway.example.com",
@@ -53,6 +56,7 @@ describe("desktop config [federation] section", () => {
     expect(config.federation).toEqual({
       mode: "dual",
       instanceLabel: "Studio Mac",
+      instanceNotes: "PwrSnap dev + screen recording",
       listenHost: "0.0.0.0",
       listenPort: 47831,
       publicUrl: "https://gateway.example.com",
@@ -67,6 +71,7 @@ describe("desktop config [federation] section", () => {
       "[federation]",
       'mode = "gateway"',
       'instance_label = "Studio Mac"',
+      'instance_notes = "PwrSnap dev + screen recording"',
       'listen_host = "127.0.0.1"',
       "listen_port = 47830",
       'public_url = "https://gateway.example.com"',
@@ -80,6 +85,7 @@ describe("desktop config [federation] section", () => {
         federation: {
           mode: "disabled",
           instanceLabel: "",
+          instanceNotes: "",
           listenHost: "",
           listenPort: 0,
           publicUrl: "",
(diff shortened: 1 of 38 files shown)


huntharo:
Follow-up round pushed (3 commits), from the review + operator discussion:

- **`scope` on `search_federation_threads`** — `all` (default) / `local` / `remote`; `remote` closes the "I know it's not on this machine" gap. Intersects with `instanceId`.
- **Host facts on every instance** — `FederationPeerSummary.host` (platform, OS version, hostname, arch, CPU count, RAM, free disk) advertised through handshake/store/gossip/health like notes/icon. `machineId` is minted once at `<pwragent root>/machine-id` and shared by all profiles on a machine, so agents can tell that "work" and "default" on one box compete for the same hardware instead of summing 16+16 CPUs. Disk figures are handshake-time snapshots.
- **Paged, filterable `list_federation_instances`** — 25 rows per page (limit 1–100), single-use ~60s continuation tokens, and a `query` substring filter over label/notes/profile/id/hostname/platform/arch so a large fleet nudges the agent toward filtering instead of consuming every row. Also fixed the search `limit` description to match the real default (50).
- Threat-model note added: peer-authored metadata (labels, notes, titles, host facts) reaches local agent context; host facts are self-reported hints under the trusted-operator boundary.

Live load signals (load averages, available RAM/disk deltas) are deliberately deferred to a spun-off task: query-on-demand RPC + Star Map health indicators + opt-in `includeLoad` on the instance list — not broadcast.

Full suite: 6341 passed. 🤖

huntharo:
Rebased onto main after #1249 (celestial instance icons) landed. Reconciliation:

- #1249 shipped the icon as a typed `FederationPeerSummary.celestialIcon` with a gateway-coordinated LWW assignment protocol — exactly the coordination point this PR reserved. This branch's free-form `icon` carry field (handshake/store/gossip plumbing) was **dropped during the rebase** in favor of that protocol; `notes` and the `host` block keep the same wire seams.
- The agent-tool descriptors now read `peer.celestialIcon` / `health.localCelestialIcon`, so agents see the same identity icons the Star Map renders. The tool contract is unchanged (descriptor field is still `icon`).
- Settings → Federation keeps both new fields: #1249's icon picker, then this PR's Purpose notes.
- Reconciliation is recorded in the plan doc ("Rebase reconciliation (post-#1249)"); note the earlier commit titled "advertise instance notes + icon" now carries notes only.

Post-rebase verification: typecheck, ESLint (0 errors), boundaries clean, full suite 6,385 passed. 🤖

PR #1304, opened 2026-08-07
fix(desktop): clear WCAG AA in light theme and gate both themes

huntharo:
## Summary

The a11y gate (`apps/desktop/e2e/a11y.spec.ts`) has only ever run in dark theme, so light-theme contrast regressions were never caught. Three light-theme tokens fail WCAG 2 AA as text — all token-level, not one-off literals.

**Root cause for the accent pair is a ramp-direction inconsistency.** The `--accent` → `--accent-strong` → `--accent-bright` ramp means increasing emphasis in both themes, but emphasis reads as *lighter* on a dark surface and *darker* on a light one. The original light palette inverted `--accent-strong` but left `--accent-bright` travelling the dark-theme direction — still lighter than `--accent` — which left ~100 `color: var(--accent-bright)` rules at 3.0–3.4:1 on *every* light surface.

Rather than mechanically darken, this runs the whole light ramp in the emphasis-as-depth direction, holding each token's hue and saturation so it stays the same tangerine and only lightness moves:

| Token | Dark (emphasis = lightness) | Light, before | Light, after |
|---|---|---|---|
| `--accent` | `#ff8a1f` | `#c45200` | `#b74c00` |
| `--accent-strong` | `#ffa33d` | `#b34a00` | `#b34a00` *(unchanged — already cleared AA, still lands between the other two)* |
| `--accent-bright` | `#ffb35c` | `#d96d00` | `#a35200` |
| `--text-muted` | `#8c857a` | `#807870` | `#736c64` |

It also changes the **measurement basis**: every value clears 4.5:1 against `--bg-panel-hover` (`#f4f0e8`), the darkest surface these can land on — not against `--bg-app`. Measuring against white alone is exactly what let `--accent` ship at 4.61:1 on white but 4.20:1 on the sidebar wordmark. Worst-case ratios are now 4.57 / 4.74 / 4.91 / 4.55.

**Gate:** the suite is parameterized over both themes (5 surfaces × 2 = 10 tests, ~33s) instead of adding parallel light-theme blocks.

## Verification

- `a11y.spec.ts` — **10/10 pass**, both themes, window-wide (no scoping, no waivers)
- **Confirmed the gate actually bites:** reverted the tokens and re-ran — 5/5 light tests fail with 10 `color-contrast` violations
- `pnpm lint:colors`, `pnpm lint:eslint` (0 errors), `pnpm typecheck`
- `theme-contract.test.tsx` (52), quit-confirmation-dialog, IntegratedTerminal — 59 tests pass
- Light-theme screenshot reviewed: wordmark, active lens tab, and the six context labels are legible; tangerine identity intact, nothing muddied toward brown

## Notes

**The reported failures understated the scope.** With the old tokens the light gate fails all five surfaces across ~24 selectors, including `.settings-nav__brand-accent`, `.settings-nav__group-label`, `.settings-number`, and the appearance toggle groups. The three originally measured were just what a thread-open scan exposed.

**`quit-confirmation-dialog.ts` hand-mirrors the light palette as literals** and uses all three as `color:`. It renders in a separate `BrowserWindow` that the renderer gate cannot reach, so it would have silently kept the failing contrast. Synced, with a comment explaining why drift there is not auto-caught.

**`runAxe` takes an optional `{ include }` scope** as requested, documented as a last resort since anything outside the scope stops being gated (prefer a `KNOWN_VIOLATIONS` entry, which waives one selector for one rule). Light theme came out clean window-wide, so **it has no call site today** — worth a reviewer's opinion on whether to keep it.

**Two deliberate omissions:**

- `.context-window-moon__disc` paints a radial gradient `bright → strong → accent` from center out. With `bright` now darkest, that 22px disc's highlight inverts (dark center, lighter rim). Purely decorative, no text, so no AA impact — but a real if subtle visual regression. Fixing it needs a light-theme rule override, which is scope beyond the token block.
- `--terminal-cursor: #d96d00` is now an orphaned literal — it was the old `--accent-bright` and no longer matches any accent token. Left as-is: it's a non-text cursor block at 3.42:1 on the terminal's white background, clear of the 3:1 non-text threshold.

Docs updated: `docs/UI-THEME.md` token tables plus a new note on ramp direction and on picking contrast against the darkest surface a token can land on.

### DIFF
--- a/apps/desktop/e2e/a11y.spec.ts
+++ b/apps/desktop/e2e/a11y.spec.ts
@@ -12,7 +12,16 @@
 // To extend: add another entry to SURFACES below, or a separate
 // `test(...)` block that drives the renderer into a state (open a
 // dialog, switch a tab) and then calls `runAxe(window)`. Launch via
-// `launchAuditApp()` so the surface is audited at rest (see below).
+// `launchAuditApp(theme)` so the surface is audited at rest (see below).
+//
+// Every surface is audited in BOTH themes. The gate ran dark-only for
+// its whole life, which is exactly why three light-theme token-level
+// contrast failures shipped unnoticed: `--accent` at 4.20:1 on the
+// sidebar wordmark, `--accent-bright` at 3.17:1 on the active lens tab,
+// `--text-muted` at 4.23:1 on the .context-grid thread-info labels.
+// Contrast is the one rule class that is genuinely theme-dependent —
+// roles, names, and focus order are not — so auditing one theme buys
+// you roughly half the coverage you think it does.
 import path from "node:path";
 import { fileURLToPath } from "node:url";
 import AxeBuilder from "@axe-core/playwright";
@@ -36,13 +45,18 @@ const specDir = path.dirname(fileURLToPath(import.meta.url));
 // state, so the gate measures real, persistent contrast instead of a
 // sub-second animation frame. No renderer JS branches on reduced motion,
 // so this only settles CSS animation — it never changes what renders.
-async function launchAuditApp() {
+async function launchAuditApp(theme: AuditTheme) {
   const app = await launchElectronApp({
     fixturePath: path.resolve(specDir, "fixtures/smoke/replay.fixture.json"),
+    appearance: { theme },
   });
   await app.window.emulateMedia({ reducedMotion: "reduce" });
   return app;
 }
+
+// Both themes ship, so both themes are gated. See the header note.
+const AUDIT_THEMES = ["dark", "light"] as const;
+type AuditTheme = (typeof AUDIT_THEMES)[number];
 
 const WCAG_AA_TAGS = [
   "wcag2a",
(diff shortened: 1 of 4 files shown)


huntharo:
## Rebased onto #1303 — heads-up on the resolution

#1303 (Star Map a11y gating) landed while this was open and overlapped substantially. Conflict was confined to `a11y.spec.ts`; `app.css` auto-merged cleanly (its reduced-motion fix for `.star-map-card-shell` and this branch's light-theme tokens don't touch).

**Kept main's shape, layered this branch's theme parameterization on top:**

- Took main's `launchAuditApp({ fixturePath, theme })` options object over this branch's positional `theme` — the Star Map blocks need `fixturePath`, and one signature carries both.
- Both branches independently added `runAxe`'s `{ include }`. Took main's, merged in this branch's guidance that it's not a tool for silencing failures.
- Wrapped the whole `describe` in `for (const theme of AUDIT_THEMES)`, so **all seven surfaces — including both Star Map blocks — are now gated in light and dark**. 14 tests, ~57s.

**One change worth a second opinion: I dropped #1303's two scoped celestial-watermark blocks.**

Their scoping had one stated reason, quoting the comment:

> The scope is what makes the LIGHT run meaningful rather than a proxy for unrelated debt: light theme is not AA-clean window-wide today (the sidebar wordmark accent lands at 4.2:1, the active lens-switch label at 3.17:1, and the context-grid `dt` labels at 4.23:1)

Those are the exact three failures this PR fixes, so the premise is now false — and `open thread view` audits that same surface unscoped in both themes, which strictly supersedes a scoped run. Leaving the blocks in would have meant keeping a narrower test justified by a comment that is no longer true.

I kept the part that was a genuine regression guard rather than a workaround — the assertion that the watermark is **actually painted** (`.thread-view__primary .celestial-watermark` → `toHaveCount(1)`), which an unscoped contrast run can't replace — by folding it into `open thread view`, with #1303's rationale preserved verbatim. @ the #1303 author: if you'd rather keep those as standalone blocks, say so and I'll restore them unscoped.

`include` now has no call site anywhere. Kept as the documented affordance with a comment recording why it exists and why nothing uses it; happy to delete it instead.

**Bonus find:** the Star Map had never been audited in light theme. It's clean — 14/14 pass.

## Verification on the merge result

- `a11y.spec.ts` — 14/14, both themes
- `star-map-thread-flow.spec.ts` (from #1303) — passes
- `lint:colors`, `lint:eslint` (0 errors), `typecheck`
- theme-contract (52) + quit-dialog + IntegratedTerminal — 59 tests
- `apps/desktop/AGENTS.md` a11y section updated; it documented the scoping as how the watermark blocks work

PR #1350, opened 2026-08-08
fix(desktop): align composer chips on the prose baseline

huntharo:
## What

Composer inline chips (`$skill`, `@file`, `@directory`, `#thread`, PR pill) sat at three different heights within the same sentence, and none of them shared the paragraph's baseline. This pins all of them to the prose baseline.

## Why they disagreed

The chips are `inline-flex`, and an inline-flex box takes its baseline from its **first flex item**. The kinds don't agree on what that item is:

| chip | first flex item | baseline it hands the line |
|---|---|---|
| `$skill`, `@file`, `@directory` | label text | the text baseline |
| `#thread` | 1em SVG icon | bottom edge of the icon (1.75px higher) |
| PR pill | 8px status dot | bottom edge of the dot (1.25px lower) |

`.composer-tiptap-input__mention` then applied one shared `vertical-align: 0.13em` to all of them — but a `vertical-align` length is measured *from the box's own baseline*, so the single nudge landed in three different places.

## Measurements

Taken against the shipping stylesheet in Chromium at the composer's real prose metrics (14px / line-height 1.6 → 22.4px line box; cap ascent 9.86px). Distances are from the paragraph baseline, positive = above.

**Before**

| chip | box top | box bottom | center vs cap-height center | line-box inflation |
|---|---|---|---|---|
| `$skill` | 16.51 | −2.39 | +2.13 high | +0.51px |
| `@file` | 16.51 | −2.39 | +2.13 high | +0.51px |
| `@directory` | 16.51 | −2.39 | +2.13 high | +0.51px |
| `#thread` | 18.26 | −0.64 | +3.88 high | +2.26px |
| PR | 15.26 | −3.64 | +0.88 high | 0 |

**3.0px spread on an 18.9px chip** (16% of chip height), and lines carrying a chip grew by up to 2.26px, so paragraph leading changed with chip kind.

**After**

| chip | box top | box bottom | center vs cap-height center | label vs prose baseline | line-box inflation |
|---|---|---|---|---|---|
| `$skill` | 14.70 | −4.20 | +0.31 | 0.00 | 0 |
| `@file` | 14.70 | −4.20 | +0.31 | 0.00 | 0 |
| `@directory` | 14.70 | −4.20 | +0.31 | 0.00 | 0 |
| `#thread` | 14.70 | −4.20 | +0.31 | 0.00 | 0 |
| PR | 14.20 | −4.70 | −0.19 | 0.00 | 0 |

Box spread **3.0px → 0.5px**, and the 0.5px remainder is IBM Plex Mono's metrics inside the PR pill holding *its* label on the baseline — the labels themselves agree exactly. Chip widths are byte-identical before and after (107.21 / 119.66 / 84.64 / 191.73 / 71.05px).

## The fix

`vertical-align: baseline` plus a zero-width-space `::before` strut, so every kind hands the line the same text baseline to align from:

```css
.composer-tiptap-input__editor .composer-tiptap-input__mention {
  vertical-align: baseline;
}
.composer-tiptap-input__editor .composer-tiptap-input__mention::before {
  content: "\200b";
  margin-inline-end: -0.35em; /* cancel the one flex gap the strut opens */
}
```

Why not a retuned `vertical-align` length: it is measured from the per-kind synthesized baseline, so any single value re-creates the spread. Why not `vertical-align: middle` (what the transcript uses): it aligns box centers, which is uniform, but parks every chip's label ~1.1px below the sentence's baseline. Aligning the labels themselves is what makes a chip read as a word in the sentence instead of a badge floating over it.

The rule is also qualified with `.composer-tiptap-input__editor` so the composer's em-based geometry outranks `.chip` (24px/12px) and `.pr-chip` (22px/11px) on **specificity** rather than on source order — today, moving `.thread-chip`'s `height: 22px` below this block would silently resize every composer chip.

## Not changed (deliberately)

- **Transcript prose** (`.thread-markdown`) measures consistently already: its chips land at −1.24 / −1.24 / −0.93 vs cap center, a 0.31px spread. Chips do change size across the send boundary (`#thread` is 18.9px/14px in the composer, 22px/12px in the transcript; a PR pill is 14px mono vs 11px mono), which is a separate sizing decision, not this alignment bug.
- **`.pr-chip--draft`'s −1px label nudge**: composer PR chips always render `pr-chip--unknown`, so the draft variant is unreachable there.

## Testing

- `apps/desktop/src/renderer/src/styles/__tests__/composer-chip-alignment.test.ts` (new) locks `vertical-align: baseline`, the strut's `content`, and the `margin-inline-end` == `-gap` invariant, so the strut isn't removed later as dead CSS. jsdom can't lay out CSS, so the layout numbers above came from a throwaway Chromium harness that loads the shipping `app.css` and the exact DOM `renderHTML` emits; the contract test is what stays.
- `pnpm test apps/desktop/src/renderer/src/styles apps/desktop/src/renderer/src/features/composer` → 454 passed.
- `pnpm lint:eslint` → 0 errors; `pnpm lint:colors` → passed.
- Not run: desktop E2E (no local runner on this machine).

### DIFF
--- a/apps/desktop/e2e/composer-chip-alignment.spec.ts
+++ b/apps/desktop/e2e/composer-chip-alignment.spec.ts
@@ -1,0 +1,221 @@
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+import { expect, test, type Locator } from "@playwright/test";
+import { launchElectronApp } from "./fixtures/electron-app";
+
+/**
+ * Measures where the composer's inline mention chips actually land, which
+ * the CSS contract test in
+ * `src/renderer/src/styles/__tests__/composer-chip-alignment.test.ts`
+ * cannot: jsdom does not lay out CSS, so that test can only assert the
+ * declarations the alignment depends on.
+ *
+ * The chips are inline-flex, and an inline-flex box takes its baseline from
+ * its FIRST flex item. The kinds disagree on what that is — the `$skill`
+ * chip leads with its label text, `#thread` leads with a 1em SVG icon — so
+ * before the `::before` strut they aligned from different origins.
+ *
+ * Measured on this fixture at this window size, with the fix and with it
+ * reverted:
+ *
+ *              chip top vs its line's text top      paragraph height
+ *   fixed      skill -0.95   thread -0.95  (0.00)   44.78 = 2 x 22.39
+ *   reverted   skill -2.77   thread -4.25  (1.48)   47.80 (+3.02)
+ *
+ * Hence the 0.5px and 1px tolerances below: roughly three times the
+ * observed spread in the passing state, and roughly a third of the
+ * regression they exist to catch.
+ *
+ * Coverage note: this drives the two autocompletes that need no fixture
+ * work, which is also the pair that was furthest apart. The PR pill (a
+ * third structure, leading with an 8px dot) is not covered — its `#`
+ * autocomplete entry only appears for a numeric query against a thread
+ * carrying an attached PR, and no E2E fixture seeds one today.
+ */
+const specDir = path.dirname(fileURLToPath(import.meta.url));
+const fixturePath = path.resolve(
+  specDir,
+  "fixtures/skill-autocomplete-interactions/replay.fixture.json",
+);
+
+/** Chips must agree this closely on where they sit within their line. */
+const OFFSET_TOLERANCE_PX = 0.5;
+/** Slack for the line-grid and chip-height comparisons. */
+const HEIGHT_TOLERANCE_PX = 1;
+
+type ChipPlacement = {
+  height: number;
+  kind: string;
+  /** Chip top edge relative to the top of its own line's text. */
+  offsetInLine: number;
+};
+
+type ComposerMetrics = {
+  height: number;
+  /** Rendered lines, counted from the paragraph's own text runs. */
+  lines: number;
+  chips: ChipPlacement[];
+};
+
+/**
+ * Line count and per-chip placement in one pass.
+ *
+ * Every reference comes from the paragraph's TEXT runs, never from a chip's
+ * own box. That distinction is the whole measurement: a chip that rides too
+ * high also drags the top of its line box up with it, so measuring a chip
+ * against its line box compares it to itself and reports zero no matter how
+ * misaligned it is. A text run sits on the baseline, so its box top is a
+ * fixed reference the chip can be wrong against.
+ */
+async function composerMetrics(paragraph: Locator): Promise<ComposerMetrics> {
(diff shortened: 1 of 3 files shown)


huntharo:
## Follow-up: review fixes + a layout test that actually measures

**Review fixes** (`0fd043639`)
- CSS comments cut from ~40 lines to 18 — the measurement narrative lives in the commit and this PR, not above the rule.
- Strut now carries empty alt text (`content: "\200b" / ""`) so it stays out of accessible names by construction rather than by observation. Geometry unchanged.
- Contract test widened to `line-height: normal` and `height: 1.35em`. Both move the chips and neither reads as alignment, so neither was covered; both assertions were mutation-checked.
- Dropped a redundant assertion and renamed the specificity test to what it actually checks.

**Layout test** (`c034144b2`) — `apps/desktop/e2e/composer-chip-alignment.spec.ts`

A golden image was considered and rejected: `visual-regression.spec.ts` hard-skips off macOS/ARM64, its PNGs need Git LFS and a Tart-lab regeneration, and a pixel diff reports "something changed" rather than the number that actually regressed. This measures instead.

Every reference is a **text run**, never a chip's own box. That distinction is the measurement: a chip riding too high drags the top of its line box up with it, so measuring a chip against its line box compares it to itself and reports zero however misaligned it is. A first cut made exactly that mistake and reported a perfect 0.00 spread on deliberately broken CSS.

Run on the PwrSuiteLab macOS Tart guest, with the fix and with it reverted — chip top vs its own line's text top, and paragraph height:

| | `$skill` | `#thread` | spread | paragraph |
|---|---|---|---|---|
| fixed | −0.95 | −0.95 | **0.00** | 44.78 = 2 × 22.39 |
| reverted | −2.77 | −4.25 | **1.48** | 47.80 (+3.02) |

Tolerances (0.5px on the spread, 1px on the line grid) are set from those two datasets — about 3× the observed spread when passing, about a third of the regression they exist to catch.

Verified on the guest, not just written:
- fix in place → **passed** (`pwragent-e2e-20260808-005900-c034144b`)
- fix reverted → **failed** on the primary assertion, `thread chip vs skill chip, top edge within its own line: 1.484 > 0.5` (`pwragent-e2e-20260808-005939-12805ea3`)

The spec also asserts the two chips genuinely lead with different elements (`$skill` has no element children, `#thread`'s first child is `<svg>`) before comparing them, so it can't pass vacuously if the icon ever stops rendering.

**Not covered:** the PR pill, a third structure leading with an 8px dot. Its `#` autocomplete entry appears only for a numeric query against a thread with an attached PR, and no E2E fixture seeds one — noted in the spec's docblock rather than left silent.

PR #1402, opened 2026-08-09
fix(desktop): stop unknown ACP updates from splitting assistant messages

huntharo:
## Problem

In `AcpSessionReplayNormalizer.apply()`, the final `else` fallback cleared `activeAssistantMessageId` / `activeAssistantMessagePhase` for **every** kind that reached it, *before* the inner if/else chain decided what the update actually was. Only `tool_call_update` for an already-known `toolCallId` was spared.

So any unrecognized session update split the assistant message being streamed in half and wedged an `ACP update: <kind>` placeholder between the halves. Probed against merged `main`, a made-up kind between two `agent_message_chunk`s produced:

```
{ type: "message",  text: "Overall: " }
{ type: "activity", text: "ACP update: some_future_grok_update" }
{ type: "message",  text: "the patch is fine." }
```

#1394 fixed two *instances* of this — `session_info_update` (routed to a metadata branch) and `last_turn_summary` (folded into `isGrokTransientUpdateKind`, also landed by #1389). Both were Grok Build, both mid-turn, and the second shipped a screenshot of a "Review summary" reply torn into two bubbles. The mechanism stayed live for the next extension update xAI adds, and for anything Gemini/Kimi/Qwen emits that is not enumerated.

## Fix

Move the bubble-clearing into the branches that have earned it rather than doing it up front:

- **Clears** — `plan`, `tool_call`, standalone `tool_call_update`, `file`, `terminal`, `turn_started`. These really are message boundaries.
- **Already cleared** — `turn_finished`, `pwragent_turn_failed`, `pwragent_user_prompt` clear via their `record*` methods; no change.
- **Preserved** — a `tool_call_update` for a known tool call keeps the bubble, exactly as before.
- **Now preserved** — an unrecognized kind falls through to `unknownActivity()` with the assistant message left intact. Failing to classify an update is not evidence that it ended the assistant's message.

The `unknownActivity` breadcrumb still lands — it is genuinely useful for spotting new protocol traffic, it just isn't destructive anymore. With the bubble left open, later chunks append to the existing (earlier) message entry, so the activity naturally sorts *after* the completed message rather than through it. That ordering consequence is commented in the code.

## Tests

Added `"keeps an unrecognized update from splitting a streaming assistant message"`, which feeds an unrecognized `sessionUpdate` between two `agent_message_chunk`s and asserts one assistant message plus a trailing activity. Confirmed it fails on the pre-fix normalizer with exactly the 3-entry split above.

The two tests pinning behavior that must NOT change stay green: `"splits assistant messages around a standalone tool update"` and `"updates known tool progress without splitting a streaming assistant message"`.

## Verification

- `pnpm test` — 6985 passed. The only failures were 5 load-dependent 5s timeouts in `acp-backend-adapter.test.ts` (known flake); that file re-run alone passes 39/39.
- `pnpm --filter @pwragent/desktop typecheck` — clean
- `pnpm lint:eslint` — 0 errors (134 pre-existing warnings)
- `pnpm lint:boundaries` — no violations

### DIFF
--- a/apps/desktop/src/main/__tests__/acp-session-normalizer.test.ts
+++ b/apps/desktop/src/main/__tests__/acp-session-normalizer.test.ts
@@ -1728,6 +1728,43 @@ describe("AcpSessionReplayNormalizer", () => {
     ]);
   });
 
+  it.each(["plan", "file", "terminal", "turn_started"])(
+    "keeps %s as an assistant message boundary",
+    (sessionUpdate) => {
+      // These kinds really do end the assistant's message. The clear that used
+      // to happen up front for everything now lives in each of their branches,
+      // so each one needs its own guard: the two tool-call tests above pin only
+      // the tool-call pair, and without this dropping any of the other calls is
+      // a silent regression.
+      const normalizer = new AcpSessionReplayNormalizer();
+
+      normalizer.apply({
+        sessionId: "session-1",
+        receivedAt: 1000,
+        update: { sessionUpdate: "agent_message_chunk", content: "Before" },
+      });
+      normalizer.apply({
+        sessionId: "session-1",
+        receivedAt: 1001,
+        update: { sessionUpdate, title: "Boundary", status: "completed" },
+      });
+      const replay = normalizer.apply({
+        sessionId: "session-1",
+        receivedAt: 1002,
+        update: { sessionUpdate: "agent_message_chunk", content: "After" },
+      });
+
+      expect(
+        replay.entries.flatMap((entry) =>
+          entry.type === "message" ? [`${entry.id}:${entry.text}`] : []
+        ),
+      ).toEqual([
+        "assistant:session-1:0:Before",
+        "assistant:session-1:1:After",
+      ]);
+    },
+  );
+
   it("treats Grok turn_completed as an idempotent turn finish", () => {
     const normalizer = new AcpSessionReplayNormalizer();
     normalizer.recordUserPrompt({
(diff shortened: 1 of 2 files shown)


huntharo:
Self-review follow-up in 6fdd12a — all three findings addressed.

**1. Three of the four boundary guards were untested.** Moving the clear out of the shared fallback turned one guarded statement into four call sites, but only the tool-call pair was pinned. Verified by mutation: deleting `endActiveAssistantMessage()` from the `plan`, `file`/`terminal`, and `turn_started` branches left all 59 tests green — a silent, complete regression of the behavior this PR is careful to preserve.

Added a table-driven test over those four kinds. Re-running the same mutation now fails all four:

```
× keeps plan as an assistant message boundary
× keeps file as an assistant message boundary
× keeps terminal as an assistant message boundary
× keeps turn_started as an assistant message boundary
```

**2. The accepted tradeoff wasn't stated.** The fallback comment now names the regression direction: if an unrecognized kind really was a boundary — a provider naming its tool call something we don't enumerate — text after it merges into the earlier bubble instead of starting a new one. That's the better failure than a reply torn in half, and the breadcrumb is what surfaces the kind to enumerate.

**3. Distinct unknown kinds collided on one activity.** `unknownActivity` built its id as `unknown:<session>:<createdAt>`, so two unrecognized kinds in the same millisecond shared an id and `upsertActivity` merged them — `preferSpecificLabel` keeps the existing summary, so the second kind's name never reached the transcript. That defeats the breadcrumb this whole fix is built around, and it gets more reachable now that unknown updates are a durable non-destructive channel. The kind is now part of the id. Two tests cover it: distinct kinds stay separate, and a re-applied update still dedupes onto one entry (the property the id exists for).

Verification after the follow-up: `pnpm test` 6996 passed / 0 failed (the `acp-backend-adapter` flake didn't recur), typecheck clean, `pnpm lint:eslint` 0 errors, `pnpm lint:boundaries` no violations.

PR #1473, opened 2026-08-10
feat(desktop): surface unsent composer drafts on thread rows

huntharo:
## Why

A half-written reply was invisible outside the thread it lived in. It persists to `composer_draft_latest` and survives a relaunch, so the text was never lost — but nothing in the sidebar, the Star Map, or any lens said it was there, and the only way to find one was to remember.

## What

**A "Draft" chip on every surface that renders a thread row**, plus a **Drafts lens** that collects them.

- `ComposerDraftStore` grows draft-presence reactivity alongside the existing queued-turn bridge. It notifies on *transitions* only — a scope gaining or losing recoverable content — not on every edit. A per-keystroke notification would re-render the whole thread list for as long as an operator is typing; a test pins that it doesn't.
- `useThreadDraftIndicators` derives the per-thread map, keyed by `buildThreadIdentityKey` so it lands on the same rows the queued and approval maps do. The Drafts lens filters from that same map, so the lens and the chips agree by construction.
- The chip is the quietest of the three pending states — a dashed neutral outline, not the accent. Scheduled and Queued are messages the app will act on; a draft is waiting on the operator and nothing happens until they come back. It survives the Compact density strip for the same reason from the other direction: it is the one pending state nothing else will ever resolve.

## Federation, remote viewers, Star Map

**A draft belongs to whoever typed it.** It stays in that window's store and that machine's sqlite, and is deliberately **not** federated:

- A federation viewer that half-writes a reply to a peer's thread sees its own chip on that row — the scope key is the same `thread:<backend>:<id>` the composer wrote it under, so the affordance works there unchanged with no wire changes, no `FEDERATION_PROTOCOL_VERSION` bump, and no new capability.
- The **owning** instance does not see it. Publishing an operator's unsent text to another machine is not worth an affordance. If we ever want a real "my draft follows me between machines" feature, that is a separate design with its own privacy call — flagging it rather than deciding it here.
- The Star Map takes draft keys as **their own prop** rather than through `sessionKeys`. The screen withholds `sessionKeys` from remote cards because thinking/approval keys come from an unscoped event stream; a draft is this window's own composer state and needs no peer to confirm it, so it applies to local and remote cards alike.

## Also in here

- **Lens switch top gap 16px → 10px.** The row above it is 6px of padding around one line of text, so 16 read as a hole and the strip floated away from the chrome it belongs to. `.sidebar__section` is only ever worn together with `--fill`, so nothing else moves.
- **`normalizeNavigationBrowseMode` moves to `@pwragent/shared`.** Three processes decoded the persisted lens independently, and preload's hand-copied allowlist had already dropped `attention` — so an operator whose saved lens was Attention got Inbox at first paint and then a visible jump, the exact flicker the bootstrap hint exists to prevent. `drafts` would have drifted the same way. Pre-existing bug, fixed in the code this PR already had to touch.
- **Four hand-rolled copies** of the `thread:<backend>:<id>` scope key collapse into `buildThreadComposerScopeKey`, and the composer's inline "is the composer empty" predicate becomes the shared `hasComposerDraftContent` the store tracks presence with.

## Validation

- `pnpm typecheck`, `pnpm lint:eslint` (0 errors), `pnpm lint:boundaries`, `pnpm lint:colors`, `pnpm lint:sql`, `pnpm lint:codex-storage` — all clean.
- `pnpm test`: 7290 passed. The 5 failures in `codex-environment-runtime.test.ts` reproduce identically on `main` — local npm config warnings leaking into captured setup output, unrelated to this change.
- New tests: draft-presence reactivity + the no-re-render-per-keystroke guarantee, the Drafts lens (membership, chip in other lenses, tab switch, tooltip, empty state), the Star Map card chip, and the shared browse-mode normalizer including the `attention` regression.

## Needs a lab run before merge

`e2e/visual-regression.spec.ts` goldens will shift — a fifth lens tab plus the tighter top gap. They have to be regenerated on the Tart lab guest, not from a desktop:

```
~/pwrdrvr/PwrSuiteLab/macos-tart/run-e2e.sh --confirm-live-run \
  --workload pwragent --local /path/to/PwrAgent \
  e2e/visual-regression.spec.ts --update-snapshots
```

The README / docs-site screenshots also show the lens strip and will drift, but those are regenerated on demand and not CI-gated.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -192,14 +192,23 @@ accidental.
 ## Current Product Direction
 
 - Threads are first-class and may exist without a directory.
-- Attention, Inbox, Recents, and Directories share the thread lens switch. The
-  tabs are icon-only; the lens name lives in `aria-label` and the tooltip.
+- Attention, Drafts, Inbox, Recents, and Directories share the thread lens
+  switch. The tabs are icon-only; the lens name lives in `aria-label` and the
+  tooltip.
 - Attention is the work-queue lens: threads with a live turn or waiting to be
   reviewed, in recent-activity order. Its tab is two indicators with counts
   (scanner + in-progress, cookie + unread) rather than an icon, and each goes
   grey at zero so the tab reads as "nothing running, nothing unread" without
   being opened. A zero is shown, never hidden — a vanishing count makes an
   idle tab look like a broken one.
+- Drafts is the second state lens: threads holding unsent composer text, in
+  recent-activity order. A draft belongs to whoever typed it — it is stored in
+  that machine's `composer_draft_latest` table, re-hydrated on the next launch,
+  and deliberately **never federated**. A viewer that half-writes a reply to a
+  peer's thread sees its own "Draft" chip on that row; the owning instance does
+  not, because publishing an operator's unsent text to another machine is not
+  something the affordance is worth. Threads carry the chip in every lens, not
+  only this one.
 - Inbox is the default browsing lens: all threads in recent-activity order.
 - Recents shows all threads in thread-creation order so active threads do not
   jump around.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -208,7 +208,11 @@ accidental.
   peer's thread sees its own "Draft" chip on that row; the owning instance does
   not, because publishing an operator's unsent text to another machine is not
   something the affordance is worth. Threads carry the chip in every lens, not
-  only this one.
+  only this one. Two limits are deliberate rather than bugs: the storage is
+  machine-wide but each window reads it once at mount, so a second open window
+  does not light up until it restarts; and launchpad (new-thread) composer text
+  has no thread row, so the lens says "replies" and its empty state says "No
+  unsent replies."
 - Inbox is the default browsing lens: all threads in recent-activity order.
 - Recents shows all threads in thread-creation order so active threads do not
   jump around.
--- a/apps/desktop/AGENTS.md
+++ b/apps/desktop/AGENTS.md
@@ -38,8 +38,11 @@ summary.
 
 ## Non-Negotiables
 
-- Attention, Inbox, Recents, and Directories live in one icon-only thread lens
-  switch; Inbox is the default browsing lens.
+- Attention, Drafts, Inbox, Recents, and Directories live in one icon-only
+  thread lens switch; Inbox is the default browsing lens.
+- A thread's unsent composer draft is window-local state, never federated.
+  Derive it from the composer draft store (`useThreadDraftIndicators`), not
+  from `NavigationThreadSummary`.
 - Attention is a work queue: focusing a thread there must not clear its unread
   cookie, only replying does. See "Current Product Direction" in the
   [repo-root `AGENTS.md`](../../AGENTS.md) for why that rule is lens-scoped
--- a/apps/desktop/AGENTS.md
+++ b/apps/desktop/AGENTS.md
@@ -151,6 +151,16 @@ elements and fail Playwright strict mode. When writing specs:
 - Target the history buttons themselves via their test ids:
   `history-nav-back` / `history-nav-forward`.
 
+The same trap runs in the other direction, and it bites the *renderer*
+rather than the spec: **`getByLabel` matches on substring**, and 31 call
+sites across 23 specs drive the composer with `getByLabel("Reply")`. A new
+`aria-label` anywhere in the app that merely *contains* "reply" turns every
+one of them into a strict-mode violation as soon as the labelled element
+renders. The unsent-draft chip was originally labelled "Unsent draft reply"
+and did exactly that — the unit suite was green and only Desktop E2E caught
+it. When you add an `aria-label`, grep the specs for the words in it before
+you settle on the wording.
+
 ## Inspecting Branch Drift Dialog E2E
 
 To open the replay-backed "Thread branch changed" dialog and keep Electron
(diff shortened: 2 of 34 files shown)


huntharo:
Pushed `de2a648` addressing the review. Summary of what changed and what I measured.

### The CI failure was real, not a flake

Playwright's `getByLabel` matches on **substring**. The chip's `aria-label="Unsent draft reply"` therefore matched the composer's `aria-label="Reply"` — used at **31 call sites across 23 specs** — turning every one into a strict-mode violation the moment a thread had a draft. `composer-image-normalization.spec.ts` hit it because a pasted image alone counts as draft content.

Relabelled to `"Unsent draft"` (one line, versus rewriting 31 locators), reproduced the failure locally, and confirmed both tests pass after the rename. Written up in `apps/desktop/AGENTS.md` next to the existing locator-hygiene section — this class bites the *renderer*, not the spec, and no unit test can see it.

### Preload bundle is back to baseline

| | `out/preload/index.cjs` |
|---|---|
| `main` | 29.25 kB |
| before this commit | 36.08 kB (**+23%**) |
| after `"sideEffects": false` | **29.27 kB** |

Preload was the first non-type consumer of `@pwragent/shared`, and with no `sideEffects` field Rollup could not shake the `export *` barrel — on a bundle parsed at every window creation, before first paint.

I verified the flag is a true claim rather than asserting it: walked the top-level statements of all 51 modules in the package. **Zero** bare statements; every flagged initializer is pure value construction (`new Set` of literals, `.map` over a local catalog, `new RegExp` from local constants, `Date.UTC` — not `Date.now`) bound to an export.

### Other findings

- **`role="img"` on the Draft chip.** On a role-less span the element is generic, where `aria-label` is prohibited and silently dropped — it would have announced as its visible "Draft" text. Fixed Scheduled and Queued the same way rather than leaving the defect I copied from them.
- **Corrected the "window-local by design" framing.** It conflated two scopes: the storage is machine-wide (`composer_draft_latest`, shared by every window on the profile); only the *view* is per-window, so a second open window stays dark until it restarts. Documented, along with the two lens gaps — threads outside the navigation snapshot (an archived thread's draft row is unreachable, and nothing deletes it) and launchpad drafts.
- **Empty state is now "No unsent replies."** — "No unsent drafts." was false whenever a launchpad draft existed.
- Nits: dropped the `getThreadComposerScopeKey` alias, memoized the `useSyncExternalStore` callbacks so the hook stops resubscribing every render, and replaced an invented "~44px per icon tab" *measurement* with the arithmetic that actually backs it.

### No new sqlite write path

Worth stating explicitly given the write-budget rules: this PR adds **no** writes. `composer_draft_latest` persistence already existed; draft-presence tracking is in-memory, and the store notifies on presence *transitions* only, so typing costs neither a write nor a re-render.

### Local E2E: what is mine and what is not

The full local sweep showed 10 failures. I ran the same five spec files on the pre-PR base commit (`91ea498d`) on the same machine to separate them:

| Failure | base | branch |
|---|---|---|
| `acp-runtime-modes` x5 | fails | fails |
| `provider-model-selectors` | fails | fails |
| `visual-regression` x2 | fails | fails |
| `app-quit-lifecycle` | passes | fails |
| `skill-autocomplete-interactions` | passes | fails |

Six are pre-existing on this desktop. The last two pass when run in isolation on this branch (8 passed, exit 0) — they were load flakes in a 200-test parallel sweep, not regressions.

**Correction to the original PR description:** I listed the `visual-regression` goldens as needing regeneration because of this change. They fail on the base commit too, so they are already stale on a developer desktop — which matches the spec header's note that only the Tart lab guest is authoritative. This change will still move them, but I could not measure that here and should not have presented those failures as evidence of it. The lab regeneration is still the right pre-merge step.

huntharo:
### macOS lane failures are repo-wide, not this PR

Everything else is green — `Desktop E2E` (the job the label collision broke), both Linux lanes, Lint, Test, and all three Windows jobs. The two macOS lanes fail in ~20s at a step that runs **before** any test:

```
bash: ./scripts/ci/macos-e2e-pre-run-cleanup.sh: No such file or directory
##[error]Process completed with exit code 127
```

"Run desktop replay-backed Electron tests" is **skipped**, so zero tests execute.

That script is tracked and has been since #1361, long before this branch's base, and this PR changes **no** file under `scripts/` or `.github/`. A full (non-sparse) `actions/checkout` reporting a missing tracked file points at the persistent Tart workspace, not at any diff.

Confirmed by looking outside this PR — same step, same error:

| Run | Branch | macOS lanes |
|---|---|---|
| 31424973512 | `main` @ 19:37 | pass |
| 31426384633 | `feat/protocol-capture-controls` @ 19:54 | pass |
| 31426940849 | `agent/codex-mcp-status-panel` @ 20:00 | **fail** |
| 31427746877 | **`main`** @ 20:10 | **fail** |
| 31427851162 | **`main`** @ 20:12 | **fail** |
| 31428343414 | `agent/ste100-tool-reflection-docs` @ 20:18 | **fail** |
| 31429623994 | this PR @ 20:33 | **fail** |

It is failing on `main`, so it is not attributable to any single PR. It broke somewhere between 19:54 and 20:00. A re-run of just the failed lanes reproduced it identically, so it is not a transient that clears itself.

**Lead, not a conclusion:** `agent/sparse-cache-checkouts` is actively adding `sparse-checkout` to these jobs and already carries a step called "Materialize full checkout after sparse cache primer" whose comment notes that `actions/checkout` disables sparse. Since the macOS runners are persistent Tart guests with shared workspaces, a leftover `.git` sparse config from one job would present exactly as "tracked file missing" in the next. That fits the re-run failing (it ran after that branch's 20:40 job) but does **not** explain the first failures at 20:00, which predate that branch's earliest run I can see. So: plausible contributor, not a proven root cause.

This is lab infrastructure rather than something I should change from here, and it needs the same lab visit as the `visual-regression` golden regeneration.

PR #1480, opened 2026-08-11
fix(desktop): apply star map card menu actions to the selection

huntharo:
## The bug

Shift-drag on the Star Map gathers cards the same way shift-click gathers rows in the thread list, and the cards stay lit when a card's kebab is opened. **Archive thread** then archived exactly one thread — the card whose kebab was used.

What the operator sees is worse than "one card archived": removing a card re-lays-out the cloud, so it visibly shifts. Reported as *"I thought 'oh it's gone!' … but it was like 6" over to the right and only the card whose kebab menu I interacted with was gone."* The gesture that selected five cards was silently discarded by the menu that came next.

## The fix

The kebab resolves its targets the way `Sidebar`'s thread context menu already does (`resolveContextMenuThreads`):

- A menu opened on a **selected** card is about the **whole selection**.
- A menu opened on a card **outside** the selection is about **that card alone**, and leaves the selection untouched.

The label carries the count — `Archive 4 threads` — so the scope is visible *before* the click rather than inferred from the aftermath. **Mark as seen** follows the same rule (over the selected cards that are actually unread). **Open in full view** stays single-target: a selection is a set of things to act on, not a set of windows to open.

Targets settle as a group via `Promise.allSettled`, so:

- one offline peer cannot cancel the archives that would have landed,
- the first failure is surfaced in the existing card-error alert,
- every owning cloud refreshes even after a partial failure, since the map is now stale for whichever cards did land.

Archived cards are removed from the selection. That is deliberately different from the existing rule for cards a filter or a flapping instance takes off the map (those stay selected on purpose — see `commitSelectionMove`); archived ones are not coming back, so a counter that keeps naming them is simply wrong.

## Tests

Five new cases in `star-map-selection.test.tsx`. Four of them fail on the pre-fix `StarMapScreen.tsx` (verified by stashing the source change); the fifth pins the unchanged single-card behavior so the fix can't over-reach:

- archives every selected card, not only the one whose kebab was used
- marks the whole selection seen from one card's kebab
- acts on one card when its kebab is opened outside the selection *(passes before and after — the guard)*
- archives the rest of the selection when one card fails, and says which
- stops counting cards it archived

`pnpm test apps/desktop/src/renderer/src/features/star-map/` — 18 files, 261 tests pass. `pnpm typecheck` and `pnpm lint:eslint` clean (0 errors, no new warnings).

## Not addressed

The cloud re-layout when cards leave is untouched — with the whole selection archived it now reads correctly.

### DIFF
--- a/apps/desktop/src/renderer/src/features/star-map/StarMapScreen.tsx
+++ b/apps/desktop/src/renderer/src/features/star-map/StarMapScreen.tsx
@@ -1113,6 +1113,19 @@ export function StarMapScreen(props: StarMapScreenProps) {
   const [marquee, setMarquee] = useState<SnapRect | undefined>(undefined);
 
   /**
+   * A card key names its owning instance, and the local instance's durable
+   * id only arrives with federation health — until then the local cloud is
+   * keyed "local". A selection swept in that window points at a cloud that
+   * no longer exists the moment health lands: the counter keeps counting,
+   * no card paints as selected, and the kebab finds nothing of it to act
+   * on. Drop it, the same way a card withholds dragging until the durable
+   * id is known rather than persisting against the placeholder.
+   */
+  useEffect(() => {
+    setSelection((current) => (current.size > 0 ? new Set() : current));
+  }, [localInstanceId]);
+
+  /**
    * Refresh whichever cloud owns a thread. Archive removes it from the
    * owning instance, so the map has to re-fetch rather than guess.
    */
--- a/apps/desktop/src/renderer/src/features/star-map/StarMapScreen.tsx
+++ b/apps/desktop/src/renderer/src/features/star-map/StarMapScreen.tsx
@@ -183,6 +183,19 @@ type StarMapCardTarget = {
   thread: NavigationThreadSummary;
 };
 
+/**
+ * What one card's menu acts on: every target, and the subset an unread
+ * action applies to. Resolved together because both are answers to the
+ * same question — which cards is this menu about — and splitting them
+ * put a per-card `filter` back in the render path.
+ */
+type StarMapCardTargets = {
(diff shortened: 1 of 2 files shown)


huntharo:
Review follow-ups pushed in e6e263ea0 — all six findings addressed.

**#1 Projects lens made an invisible selection destructive (the real bug).** Two holes, both closed:
- `cardRects` is now empty under the projects lens. That lens draws none of `bodies` — its cards sit on project arms — so a marquee there was measured against lane rects nothing is painted at, selecting cards the operator could not see.
- A lens change drops the selection, so one swept in Lanes can't be carried into a lens that paints no selected state.

Both are pinned by new tests, and both were confirmed failing with the guards reverted.

**#2 Failure copy carries the counts.** `Could not archive 1 of 3 threads: peer is offline`. Single-card actions read `Could not archive that thread: <reason>` — which also fixes the old behavior where the reason replaced the context entirely.

**#3 Prune moved into the per-card success path.** A card whose archive was refused is still on the map, so it stays selected; only the ones that actually left drop out of the count.

**#4 Reverted the error auto-clear.** A successful action no longer wipes an unrelated error nobody has read; the dismiss button is still the way out. (This was an undeclared behavior change in the first commit.)

**#5 Selection resolved once per render** via a memo, instead of once per card — the render was quadratic in a selection an operator can sweep to 40 cards a cloud.

**#6 Documented** the deliberate mark-seen subset choice in a comment.

Test gaps from the review are filled too: refresh-once-per-cloud (dedupe), selected cards no longer on the map being skipped, and the retained-selection-after-failure case.

265 star-map tests pass; full renderer suite 2433/2433; typecheck clean; ESLint 0 errors, no new warnings.

PR #1492, opened 2026-08-11
fix(desktop): name remote threads in the quit dialog

huntharo:
## The report

Quitting with a peer's integrated terminal running showed the blocker row as a raw UUID — while the viewer window right behind the dialog was already showing that same thread by name.

## Why

`quit-manager`'s `resolveThreadTitles` asked `BackendRegistry.listThreads({ callerReason: "quit-confirmation" })` — **this instance's** threads. A federated thread isn't in that list, the lookup returned nothing, and the row fell back to the thread id.

## Three tiers of thread-name resolution

Named on `RemoteThreadSummaryCache` so the choice is explicit rather than incidental:

1. **Local only** — this instance's thread list. Correct only when the caller knows every thread it can see is its own. This is what the quit dialog was doing.
2. **Local + cached remote** — `cachedThreadNameFromPeer`. A map read: no fetch, no TTL-triggered refetch, no peer-interest registration. The right default, and what the quit dialog now uses.
3. **Local + live remote** — the existing `threadFromPeer`, for callers that genuinely need current data and can afford to wait on a peer.

Reaching the peer is deliberately **not** done on the quit path. A remote thread is a blocker because a window on this machine has its terminal open, so some snapshot named it on the way to the screen — a round trip would re-answer a question we can answer while making shutdown wait on a machine that may be asleep. No knob is added for tier 3 here: it has no caller on this path, and a parameter with one reachable value would be dead code.

### Where remembered names come from

`RemoteThreadSummaryCache` is populated only by pinned-thread merges and ⌘K jump search — a **federation window** reads its navigation straight from the runtime and never touches that cache. Naming off that cache alone would have left the federation-window case still showing a UUID.

So the cache keeps a **name index**, merged from every snapshot that reaches this instance, including the federation-window navigation read in `readNavigationSnapshot`. Design notes:

- **Names only, not summaries.** This exists so a display string is in hand; holding whole rows for every thread ever seen on every peer would cost real memory for no extra answer.
- **Merged, never replaced.** Snapshots are backend- and filter-scoped, and a narrower one must not erase what a wider one taught us.
- **Fallback titles skipped.** A `titleSource: "fallback"` title *is* the thread id — recording one would overwrite a real name with nothing.
- **Survives TTL lapse, disconnect, and `invalidate`.** A name one refresh out of date beats a UUID, and every alternative costs the round trip tier 2 avoids.
- **Never fails a snapshot.** Remembering is a side errand, wrapped so a broken federation runtime can't break the navigation the operator actually asked for.

Second cached source: the pinned rows' persisted summaries, which survive a restart so a freshly launched app can still name a pinned remote thread before anything has talked to that peer. Neither source is a guarantee — a peer thread no snapshot has ever named still falls back to its thread id, which is the correct outcome.

## Audit: two more peer-blind lookups behind the same rows

Both on the click path of the row the operator is being asked to decide about:

| Lookup | Symptom |
|---|---|
| `revealIntegratedTerminal` consulted only the local PTY registry | Clicking a **remote** terminal row never opened its terminal panel — the local registry reports "no such session" for every shell running on a peer. |
| `requestShowThread` took only `{backend, threadId}` | The dialog holds focus at that moment, so the request fell through to `subscribers[0]`. For a federation window's terminal that's a window which never mounted the thread. |

Turns and environment actions are driven by this instance's registry and runtime, so they're local by construction. `requestShowThread`'s other caller (native approval notifications) is likewise local-only.

## Other changes

- **Quit snapshot carries the owner.** `IntegratedTerminalQuitSnapshot.threadKeys: string[]` → `threads: IntegratedTerminalQuitThread[]`, each optionally carrying its `FederationRemoteTarget` and the peer's display label.
- **Titles keyed by instance + thread key**, so a remote row can never inherit a same-keyed local thread's name — and the **row action carries the owning instance** for the same reason, so the click path can't reveal a local shell for a peer's row.
- **`revealIntegratedTerminal` reports an owner even when two windows host the same remote thread.** Remote mounts allocate a PTY per viewer window, so that is reachable; any owner demonstrably has the thread, which is more than the focused-or-first fallback can promise.
- **Deterministic sort order.** Thread keys are dense with `:` and `-`, so `localeCompare` would make row order depend on the host locale; a shared comparator restores code-unit order.
- Remote rows show the owning peer's name as their detail line; without it a remote shell reads exactly like a local one.

**PTY allocation is untouched** — `createOrAttach`, `pty.open`, `write`, `resize`, `close`, `dispose` all unmodified. Remote mounts keep their isolated per-window PTY with no replication or dual-send; the new `revealSession` only flips `panelHidden`.

## Tests

- `remote-thread-summary-cache.test.ts` — remembered names answer without contacting the peer (cold and warm); names handed in from outside the cache are remembered; merge doesn't erase and fallback titles are ignored; names survive TTL lapse and `invalidate`.
- `quit-manager.test.ts` — remote title from remembered names; pin fallback when nothing is in memory; never waits on a peer (a dependency that never settles can't block resolution); `titleSource: "fallback"` ignored; a remote thread's title kept off a same-keyed local thread; local rows still named when the remote lookups throw; snapshot carries target + label.
- `app-server-ipc.test.ts` — the federation-window read remembers names, and a throwing runtime still serves the snapshot.
- `federation-terminal-ipc.test.ts` — quit snapshot carries target + label; remote reveal reaches the owning window and only that window; an owner is named when two windows host the same thread; naming an instance skips the local registry.
- `quit-confirmation-dialog.test.ts` — the action round-trips the owning instance (and local rows don't grow one); row click routes show-thread to the reveal's owner.
- `quit-dialog-render.test.ts` — remote row renders the thread name and peer label, not the UUID.

Full suite: 521/523 files. The two failures are pre-existing and environmental, both confirmed failing identically on the base commit — `codex-environment-runtime.test.ts` (local `npm` config warnings leak into captured command output) and `acp-backend-adapter.test.ts` (5s per-test timeouts under full-suite load; passes in isolation, and imports nothing this PR touches). `typecheck`, `lint:eslint` (0 errors), `lint:boundaries`, `lint:sql`, `lint:codex-storage` all clean.

### DIFF
--- a/apps/desktop/src/main/__tests__/app-server-ipc.test.ts
+++ b/apps/desktop/src/main/__tests__/app-server-ipc.test.ts
@@ -87,6 +87,7 @@ const federationMock = vi.hoisted(() => {
     ),
     searchForJump: vi.fn(async () => ({ results: [] })),
     threadFromPeer: vi.fn(async (): Promise<unknown> => undefined),
+    rememberThreadNames: vi.fn(),
   };
   return {
     remoteBackend,
@@ -1815,6 +1816,52 @@ describe("app server ipc", () => {
         federation: { peerStatus: "disconnected" },
       }],
     });
+  });
+
+  // Remote thread names are remembered here because a federation window's
+  // navigation never reaches RemoteThreadSummaryCache. That is a side errand:
+  // the operator asked for a snapshot, and losing a name costs a thread id in
+  // a quit dialog somewhere, not this window's contents.
+  it("remembers remote thread names without letting that break the snapshot", async () => {
+    const { registerAppServerIpcHandlers } = await import("../ipc/app-server");
+    const { NAVIGATION_SNAPSHOT_CHANNEL } = await import("../../shared/ipc");
+    const federationTarget = {
+      scope: "remote" as const,
+      instanceId: "peer_remembers_names",
+    };
+    const snapshot = {
+      backend: "all" as const,
+      fetchedAt: 1_000,
+      federationTarget,
+      unchanged: false,
+      threads: [],
+      inboxThreadKeys: [],
+      directories: [],
+      launchpadDefaults: {
+        backend: "codex" as const,
+        executionMode: "default" as const,
+      },
+    };
+    federationMock.runtime.remoteNavigationSnapshot
+      .mockResolvedValueOnce(snapshot)
+      .mockResolvedValueOnce(snapshot);
+
+    registerAppServerIpcHandlers();
+    const handler = handlers.get(NAVIGATION_SNAPSHOT_CHANNEL);
+
+    await expect(handler?.({}, { federationTarget })).resolves.toBe(snapshot);
+    expect(
+      federationMock.remoteThreadSummaries.rememberThreadNames,
+    ).toHaveBeenCalledWith("peer_remembers_names", snapshot.threads);
+
+    federationMock.remoteThreadSummaries.rememberThreadNames.mockImplementationOnce(
+      () => {
+        throw new Error("federation runtime unavailable");
+      },
+    );
+    await expect(
+      handler?.({}, { federationTarget, forceRefresh: true, refreshMode: "full" }),
+    ).resolves.toBe(snapshot);
   });
 
   it("overlays viewer-owned directory disclosure state on remote snapshots", async () => {
--- a/apps/desktop/src/main/__tests__/federation-terminal-ipc.test.ts
+++ b/apps/desktop/src/main/__tests__/federation-terminal-ipc.test.ts
@@ -76,7 +76,12 @@ vi.mock("electron", () => ({
   },
 }));
 
-vi.mock("../terminal/integrated-terminal-service", () => ({
+vi.mock("../terminal/integrated-terminal-service", async (importOriginal) => ({
+  // byThreadKey is a pure comparator the merge under test uses; only the
+  // service itself needs faking.
+  ...(await importOriginal<
+    typeof import("../terminal/integrated-terminal-service")
+  >()),
   IntegratedTerminalService: class {
     constructor(options: { onSessionsChanged: (sessions: unknown[]) => void }) {
       mocks.localSessionsChanged = options.onSessionsChanged;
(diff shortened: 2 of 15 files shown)


huntharo:
Pushed 87b8a11 in response to review feedback.

## The blocking call was the wrong tier

The first pass reached the peer (`threadFromPeer`) bounded by a 750ms timeout. That was wrong, and the reasoning that motivated the timeout was itself the tell: a thread can only be a quit blocker **because it is mounted in a window on this machine**, which means its summary was already fetched and cached. The name is in hand. A round trip could only re-answer a question we can already answer, while making shutdown wait on a machine that may be asleep.

Names now come from two local reads only — the in-memory peer navigation cache and the pinned rows' persisted summaries — and the store read is skipped entirely when the memory cache already named everything. `settleWithin` and `QUIT_PEER_TITLE_TIMEOUT_MS` are gone; nothing on this path can wait on a peer.

## Three tiers, named

Documented on `RemoteThreadSummaryCache` so the choice is explicit rather than incidental:

1. **Local only** — this instance's thread list. Correct only when the caller knows every thread it can see is its own. This is what the quit dialog was doing.
2. **Local + cached remote** — the new `cachedThreadFromPeer`. A map read: no fetch, no TTL-triggered refetch, no peer-interest registration. The right default, and what the quit dialog now uses.
3. **Local + live remote** — the existing `threadFromPeer`, for callers that genuinely need current data and can afford to wait.

Tier 2 deliberately ignores the TTL *and* the peer's connection state: a name one navigation refresh out of date beats the raw thread id, and both alternatives cost the round trip the tier exists to avoid. Pinned summaries are persisted, so a freshly launched app can still name a pinned remote thread before anything has fetched from that peer.

No knob was added for tier 3 on this path — it has no caller here, and a parameter with one reachable value would be dead code. Say the word if you'd rather have the switch now.

## PTY allocation is untouched

Confirmed by diff — remote mounts still open their own isolated PTY per viewer window, and nothing replicates or dual-sends output:

- `createOrAttach`, `pty.open`, `write`, `resize`, `close`, `dispose` — **unmodified**.
- `quitSnapshotSessions` — read-only; adds `target` + `instanceLabel` to what it already returned.
- New `FederationTerminalBridge.revealSession` — flips `panelHidden` and re-broadcasts the session list. That is panel *visibility*, the same thing the local service's `revealSession` has always done. It creates and destroys nothing.

## New tests

- `remote-thread-summary-cache.test.ts` — `cachedThreadFromPeer` answers from cache and never calls `fetchSnapshot` (cold or warm); serves a lapsed entry rather than refetching.
- `quit-manager.test.ts` — "never waits on a peer" (a dependency that never settles cannot block resolution); the store read is skipped when the memory cache answered.
- Replaced the now-obsolete "waiting out a slow peer" test, which only existed to justify the timeout.

Full suite unchanged: 522/523 files, same 5 pre-existing environmental failures in `codex-environment-runtime.test.ts`. Typecheck, ESLint (0 errors), and boundaries clean.

PR #1538, opened 2026-08-12
feat(desktop): add managed MCP gateway

huntharo:
## Summary

- add a profile-scoped PwrAgent registry for remote MCP connections
- centralize OAuth discovery, authorization, refresh, and encrypted credential persistence in the profile owner process
- proxy managed MCP sessions to Codex and ACP threads through authenticated local bridge grants
- add Settings management and per-thread selection UI for generic managed MCP connections while preserving PwrSnap compatibility
- add network-safety, ownership, refresh-concurrency, lifecycle, and SQLite write-budget coverage

## Why

Codex currently owns OAuth state independently for each MCP registration. Datadog and Atlassian refresh attempts can fail with `invalid_grant` or `unauthorized_client`, forcing the operator through browser authorization again even though PwrAgent already has a long-lived parent process and a local MCP proxy pattern.

This change makes PwrAgent the single credential and refresh authority for each profile. Threads receive short-lived local bridge grants, while only the profile owner process can read or rotate provider credentials. Concurrent refresh demand is collapsed into one transaction, rotated refresh tokens are persisted before publication, and secondary PwrAgent processes forward lifecycle work to the elected owner.

## Operator impact

Operators add a remote MCP endpoint under **Settings → Plugins**, authorize it once, and choose which connections a thread may use. Provider access and refresh tokens remain encrypted in PwrAgent secret storage and are never projected into thread configuration or bridge environment variables.

Live Datadog and Rovo qualification still requires an interactive operator consent run. Later expansion beyond the current tool-oriented MCP bridge is tracked in the plan artifact.

## Product controls

The first implementation persisted more state than it exposed. These commits close that gap.

**MCP access panel (`McpAccessPanel`).** Selection opens from the composer's thread options, beside the other per-thread execution settings, and replaces the two stacked billboards that previously sat above a new thread's composer. It works on an existing thread, not only a launchpad.

**Honest apply timing.** Codex re-reads the selection while starting a turn, so a change applies to the next message. ACP resolves MCP servers only at `session/new` and `session/load`, so there it waits for the next session load. The panel states whichever applies instead of reporting a bare "saved".

**Remediation instead of a dead control.** A connection that needs authorization keeps its row and gets an **Authorize** action routing to Settings → Plugins. Previously it rendered as a disabled switch that named a problem and withheld every way to act on it.

**PwrSnap's setup ladder in Settings.** PwrSnap is a local application, not a remote OAuth endpoint, so its row offers Get / Open / Connect / Reauthorize as the state requires. A bare Authorize before the app is installed and running with Local Agent Access would fail with a connection error and name no cause.

**Availability, exposed.** `enabled` was written to TOML and enforced in the backend but had no control anywhere. Settings → Plugins now carries a per-connection **Offer to threads** switch and a profile-wide **Managed MCP gateway** switch. Parking a connection keeps its credentials and authorization state; only `remove` discards them, and the gateway switch restores every selection when turned back on.

**Provider precedence.** A Codex thread can turn off the agent's own MCP servers, expressed as `{ enabled: false }` overrides for servers Codex actually reported. When the installed Codex build cannot report its inventory, the turn is refused rather than run with the operator's other servers still live behind a UI claiming otherwise. Only Codex can honor this, so the control appears only there.

**Correct attribution on refusal.** `registerBridge` now checks availability before credentials. A connection withheld by either switch is not an authorization problem, and the old order told the operator to reauthorize — sending them to the wrong control.

**Naming.** `PwrSnapConnectionService` → `McpConnectionGatewayService`. The service stopped being PwrSnap-specific once it brokered every managed connection.

The remote PwrSnap card is deliberately kept: it is a thread-specific offer from the machine that owns the thread, and the local panel refuses to edit a remote thread's selection.

## Validation

- full workspace suite: 595 files, 8472 passed, 5 skipped
- new coverage: `McpAccessPanel.test.tsx` (7), `codex-mcp-isolation.test.ts` (3), registry parking (3), gateway availability refusal (1), Settings gateway + availability switches (1)
- `pnpm typecheck`
- `pnpm lint:eslint` (0 errors; existing warning baseline only)
- `pnpm lint:boundaries`
- `pnpm lint:colors`
- `pnpm lint:codex-storage`
- panel geometry and WCAG AA contrast measured in a headless harness across both themes (lowest ratio 5.21:1; no horizontal overflow; no sub-24px targets)

## Plan

See `docs/plans/2026-08-12-001-feat-pwragent-managed-mcp-connection-gateway-plan.md`.


huntharo:
## Settings → Plugins, before and after the UX review

Both renders come from a headless harness driving the real renderer surface against a **fully contrived fixture** — invented connections (Datadog, Atlassian), an invented Codex profile at `/home/example/.codex/profiles/work`, invented servers. No operator data.

### Before

MCP connections settings before the review

Four problems visible in that one screen:

1. **PwrSnap and PwrGit render twice**, once per section, with two different vocabularies for the same fact — `Not authorized in PwrAgent` above, `Not connected` below.
2. **`On` beside `Not connected`.** Four independent switches (gateway, per-connection `enabled`, credentials, per-thread selection) collapsed onto one row with no precedence, so the row could contradict itself. `enabled` defaults to `true` for a connection that has never been authorized, which is how that pairing happens on a fresh profile.
3. **The switch column is ragged** — each row is its own CSS grid, so an `auto` column resolves against that row's own content. The four "Offer to threads" switches land at four different x positions.
4. **No statement of what can be added.** `authMode` is the literal `"oauth"`, so a command-line (stdio) server can never be registered here — but nothing said so until discovery failed after the record was written.

### After

MCP connections settings after the review

- **One list.** The PwrSuite section is gone; PwrSnap and PwrGit are rows in the managed list like anything else.
- **One claim per row,** resolved by `resolveMcpConnectionSetup` in `packages/shared` with an explicit precedence order (gateway off → not installed → not running → not set up → login required → unavailable → connecting → parked → ready). The chip is the state; the line under it is the remedy. The availability switch is withheld entirely when there are no credentials to park, so `On`/`Not connected` is unrepresentable.
- **Aligned columns.** Availability moved to a fixed trailing column; all four switches and all four action groups share an edge.
- **Probe before write.** The add button reads *Check* until the URL answers, then *Add and authorize*. A stdio command line is refused by name rather than by an OAuth discovery failure.
- **The header chip counts readiness** (`1 ready · 1 parked · 2 to set up`) instead of counting switch positions.

Issue #240, opened 2026-05-08
ux(desktop): collapse double margins between transcript box and composer (outer + inner padding stacking)


huntharo:
## Summary

The chat transcript area has a visible box (border / background) wrapping the transcript content. The composer (reply input) sits below it with its own outer margin AND its own inner padding. The result is **stacked spacing** — the transcript's bottom padding + the gap between transcript and composer + the composer's outer margin + the composer's inner padding all line up to produce a noticeably-too-wide visual gutter, and a similar effect on the sides where the transcript's inner padding doubles up against the composer's outer margin.

This is a layout-tightening pass: reduce the redundant spacing so the two surfaces feel like one connected pane rather than two boxes wedged near each other.

## Goal

The transcript and composer should read as **one continuous chat surface**, not two boxed regions. Concretely:

- The bottom edge of the transcript and the top edge of the composer should share their gutter — one space token between them, not two.
- The horizontal alignment of message rows in the transcript should match the horizontal extent of the composer's text input (or be intentionally inset, but not "almost-but-not-quite-aligned").
- The pane background should appear contiguous; the transcript "box" treatment (if any) should not visually fight the composer's own surface treatment.

## Why

- The wasted vertical space pushes the composer further from the last message, which makes the action-context relationship feel disconnected.
- Inconsistent horizontal insets (e.g., the last message ends at one X position, the composer text starts at a slightly different X position) is the kind of detail that doesn't register consciously but makes the surface feel uncared-for.
- Once new surfaces start landing in this zone (the queued-message pill from [#237](https://github.com/pwrdrvr/PwrAgent/issues/237), the queued-permission-mode banner, attachment thumbnail strips), every extra gutter compounds. Tighten now, before more elements land in the gap.

## Acceptance criteria

- [ ] **Audit the spacing tokens** in [\`apps/desktop/src/renderer/src/styles/app.css\`](apps/desktop/src/renderer/src/styles/app.css) for the transcript pane, the composer wrapper, and any intermediate container. Identify every \`padding\`, \`margin\`, \`gap\` value contributing to the vertical gutter between transcript bottom and composer top, and to the horizontal inset on either side.
- [ ] **Pick one source of spacing per direction.** Either the transcript pane defines the bottom gutter and the composer has \`margin-top: 0\`, or the composer defines the top gutter and the transcript has \`padding-bottom: 0\`. Same call horizontally — pick a single inset and let the other surface inherit / align to it.
[...]

PR #1570, opened 2026-08-13
feat(desktop): tighten left-sidebar density, typography, and disclosure chrome

huntharo:
## Summary

Design-review pass over the thread-list sidebar, from the "left bar is too chunky" feedback. Reviewed live against the dev app with a structured critique (hierarchy / consistency / usability), then implemented the tightening.

### Design review findings → fixes

| Finding | Fix |
|---|---|
| Directory headers and thread titles shared one heavy style (15px/650) — container and content had identical weight, so the tree read flat and loud. (The style guide itself specifies smaller/lighter row text; 15/650 was an issue #240-era deviation.) | Thread titles **13px/600** by default; directory labels ride **1px smaller in `--text-secondary`**. New **Sidebar text size** setting (XS–XL) moves titles ±2px — the old 15px is the XL notch. |
| Two disclosure vocabularies: border-stroke chevron (directory rows, "Directory threads" divider) vs. filled triangle (sub-thread toggle, sub-agents). | One vocabulary: the **filled caret** everywhere in the sidebar. |
| `border-top` hairline between every directory row double-encoded a boundary the 32px headers already carry. | Hairlines removed. |
| Folder icon on every plain directory row carried zero information. | Icon only where it distinguishes: workspace + unlinked keep their marks. |
| Launchpad `+` always visible on every header while the row kebab is hover-only. | `+` is **hover/focus-revealed**, and stays pinned while it carries a draft (there the icon *is* the signal). |
| 24px chip pills under the titles were the bulkiest thing on the card (~120px for a pinned worktree thread). | Sidebar-scoped shrink to **20px/11px** — the shared `.chip` primitive, thread-header eyebrow chips, and transcript chips are untouched. Row padding 8→6px. |
| Unread cookie was 12px in rows vs 9px beside directory counts, sat ~1px below the text's optical center, and the title gap was too wide. | One **10px** cookie everywhere, nudged onto the title line, heading gap 8→6px. |
| **Bug:** pressing a card buried its focus ring's edge under the previously selected card below (`.is-selected` z-index 1 vs. unlifted focused row). | Focused card lifts to **z-index 2**. |
| **Bug:** in Compact, a pinned thread's pin chip kept a whole second chip row alive — exactly the space Compact exists to reclaim. | Compact suppresses the pin chip and renders an **in-title pin marker** instead (non-interactive by necessity — it lives inside the open-thread button; unpin stays on the context menu). |
| Settings copy claimed Compact hides PR chips; it never did (`.pr-chip` isn't matched by the density strip). | "Density" → **"Info density"**, copy now matches actual behavior. |

### New setting plumbing

`[general.appearance] sidebar_text_size = "xs" | "sm" | "md" | "lg" | "xl"` (default `md`, defaults not written to disk), flowing through the existing no-flash appearance path: shared contract → desktop-config TOML read/write → settings service (+ `onAppearanceChange` broadcast to aux windows) → bootstrap args → preload → `index.html` inline script → `useAppearance` → `<html data-sidebar-text>`. Additive config change only.

## Testing

- `pnpm typecheck`, `pnpm lint:eslint` (0 errors), `pnpm lint:colors`, `pnpm lint:boundaries` — all pass
- Targeted vitest: shared settings contract, desktop-settings-service (round-trip extended to the new key), screenshot-appearance, profiles-ipc, settings-screen, app-shell, theme-contract, navigation suites — all pass
- Verified live in the dev app (both densities, all five text-size notches, hover reveals, pressed-ring overlap repro before/after, compact pin marker)

## Follow-ups (deliberately out of scope)

- **Info-density presets as per-field defaults** (Mission Control / Balanced / Compact / Custom picking provider, location indicator, directory count caps, branch chips, PRs, emojis, pin placement per-field)
- **Balanced preset** (branch chips off, pin on title row)
- **PR chip placement option** (bin-pack after provider/location instead of always-last)
- Pin placement as an explicit setting (Hide / Title bar / Chip bar)

### DIFF
--- a/apps/desktop/e2e/directory-launchpad-workspace.spec.ts
+++ b/apps/desktop/e2e/directory-launchpad-workspace.spec.ts
@@ -1078,9 +1078,16 @@ test("directory launchpad keeps new worktree as the sticky default after startin
       .first();
     // Renamed in the chip-flow refactor: meta + PR + binding + reactions
     // chips all share a single `.thread-row__chips` flex-wrap container.
+    // The location chip is icon-only since the 2026-08 density pass, so the
+    // worktree/local kind lives in the copyable chip's aria-label rather
+    // than as visible chip text.
     const startedThreadChips = startedThreadRow.locator(".thread-row__chips");
-    await expect(startedThreadChips.getByText("worktree", { exact: true })).toBeVisible();
-    await expect(startedThreadChips.getByText("local", { exact: true })).toHaveCount(0);
+    await expect(
+      startedThreadChips.getByLabel(/^Copy path for worktree /),
+    ).toBeVisible();
+    await expect(
+      startedThreadChips.getByLabel(/^Copy local path for /),
+    ).toHaveCount(0);
 
     const contextRail = app.window.getByRole("complementary", {
       name: "Thread context",
--- a/apps/desktop/e2e/federation-remote-window.spec.ts
+++ b/apps/desktop/e2e/federation-remote-window.spec.ts
@@ -1233,9 +1233,12 @@ test.describe("federation remote window", () => {
         timeout: 90_000,
       });
 
+      // Anchored regex, not `exact: true`: this test pins the parent, and
+      // a pinned row's accessible name is "<title>, pinned" — while the
+      // anchor still excludes the row's kebab/sub-thread labels that
+      // merely contain the title.
       const remoteParentButton = viewer.window.getByRole("button", {
-        name: "Remote Kimi parent",
-        exact: true,
+        name: /^Remote Kimi parent(, pinned)?$/,
       });
       await remoteParentButton.click({ button: "right" });
       await expect(
--- a/apps/desktop/e2e/visual-regression.spec.ts-snapshots/todo-thread-shell-darwin.png
+++ b/apps/desktop/e2e/visual-regression.spec.ts-snapshots/todo-thread-shell-darwin.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:2618e70e464adc7fc82dc0485787008c9ac0dca1afa48f21203ff1c620cb4ced
-size 148993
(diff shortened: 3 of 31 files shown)


huntharo:
**Round two** (from live review feedback):

- Cookie→title spacing cut to parity with the directory-header counts (heading gap 8→2px; one 10px cookie everywhere).
- `worktree`/`local` chips are now icon-only, and **PR chips pack before the branch chip** (provider → location glyph → PRs → branch) — typical rows drop a full chip line.
- Directory launchpad `+` now overlays the header's right edge *inside* the hover border and **swaps with the count block on hover** (same pattern as timestamp ↔ kebab). It keeps `pointer-events` at rest so Playwright's pre-move hit-target check passes — this is also the fix for the launchpad E2E timeouts in the first CI run.
- **Pin left the chip flow entirely**: pinned state is the in-title marker (both densities), and pin/unpin is a hover-revealed one-click toggle beside the reaction + kebab buttons.
- `chevron-contract.test.ts` updated deliberately: the sidebar (thread tree + directory header + directory-threads divider) is now one FILLED-triangle family; unfilled −45/45 stays the inline-content-disclosure language.

huntharo:
**Round three + rebase:**

- Inter-card spacing tightened to ~10px text-to-text (row block padding 6→4px, list gap 4→2px, both densities) — ~62% of the previous blank band, per review. The focused ring intentionally paints across the 2px seam; the z-index lift keeps it on top.
- Rebased onto main over #1522/#1562. The launchpad split cluster (+ / federation-target chevron) from #1562 is preserved and now rides the hover-reveal overlay as one unit: the whole cluster swaps with the count block on hover, `has-draft` keeps it pinned visible with counts shifted beside it (wider shift when split).
- Branch history squashed to one commit as part of the rebase (three conflict passes over the same hunks otherwise).
- The `theme-contract` pin-boundary test from main updated for the shared 2px gap + `-2px` compensation.

PR #1734, opened 2026-08-17
fix(desktop): launch Windows Codex through its .cmd shim

huntharo:
## Summary

- resolve a Windows Codex command to the sibling Windows can actually start (`codex.exe`, then `codex.cmd`) instead of the extensionless sh shim or the `.ps1`
- apply that resolution at discovery **and** at every Codex spawn site, so Settings, the app-server transport, the connection test, and Codex login all agree on one command
- stop rendering raw spawn errors as status chips, which made the Codex panel unreadable
- drop the PowerShell and AppX/MSIX discovery paths added earlier in this branch

## Root cause

Two independent bugs produced one confusing symptom: Settings reported Codex **Connected** while starting a thread failed with `json-rpc timeout: initialize`.

**Discovery picked an unusable file.** `@pwrdrvr/codex-discovery` walks `[command, ...PATHEXT]` in that order. npm installs three shims side by side — `codex` (an sh script), `codex.cmd`, and `codex.ps1` — so in an nvm-windows bin directory the scan stops on the sh shim, which Windows cannot execute, and never reaches `codex.cmd`. A working install reported `Missing`.

**The `.ps1` cannot host the app-server.** npm's shim branches on `$MyInvocation.ExpectingInput`. A one-shot `--version` closes stdin and takes the direct branch, so it probes clean. The transport holds stdin open as a pipe, so it takes `$input | & node …`, routing a bidirectional JSON-RPC stream through PowerShell's object pipeline. Measured on the Windows lab guest, spawning `app-server` with stdin open and sending `initialize`:

| launch surface | result |
|---|---|
| `powershell -File codex.ps1` | **timeout after 15s** |
| `cmd.exe /c codex.cmd` | handshake OK in 1583ms |
| native vendor `codex.exe` | handshake OK in 303ms |

So `.ps1` is a discovery signal, never a launch target. `.cmd` already routes through `createCommandInvocation`'s cmd.exe wrapper, which is the standard Node-on-Windows path.

The MSIX/AppX discovery is removed because it cannot work: there is no `WindowsApps` alias on the guest, and `C:\Program Files\WindowsApps\…\codex.exe` is `EPERM` by ACL for anything but the packaged app. It only ever produced a permanently-red row plus a 5s PowerShell spawn per discovery.

## Settings panel

Discovery failures arrive in two shapes: short tokens like `not_found`, and raw spawn errors that are whole command lines. `describeCommandDiscoveryFailure` ended in `return reason`, so one landed in a chip — twice, as both the version and the status. Chips are `flex: 0 0 auto`, so the row overflowed and the path beside it collapsed to `C:`. Failures now classify to chip-sized labels with the raw text on a wrapping detail line, and `GitSettings` shares the same module rather than its own copy of the flaw.

## Validation

- `pnpm typecheck`, `pnpm lint:eslint` (0 errors), `pnpm lint:boundaries`, `pnpm lint:colors`, `pnpm lint:codex-storage`
- unit suites for `codex-windows-launch`, `codex-discovery-coordinator`, `stdio-transport`, `credential-tester`, `index`, and `settings-screen`
- `e2e/windows-codex-discovery.spec.ts` on the PwrSuiteLab Windows guest: 2/2 passing. The new case plants the real npm shim trio and asserts discovery lands on `codex.cmd` and the `.ps1` is never invoked.

**Coverage limits, stated plainly:**

- Playwright runs on Ubuntu and the self-hosted macOS group only. There is no Windows Playwright lane, so `windows-codex-discovery.spec.ts` executes only when someone drives the lab by hand.
- The E2E fixture plants its shims in a temp directory, so it does not cover a machine that also has `%APPDATA%\npm\codex.cmd` present — the plain npm-global layout.

## Follow-ups

- The upstream ordering bug is the real defect and is not fixed here: `buildPathCommandNames` returning the bare name first is wrong for win32 generally, since `CreateProcess` cannot execute an extensionless file. It affects `git` and `gh` discovery too. Tracked for `pwrdrvr/agent-kit`.
- `@pwrdrvr/codex-discovery` probes `--version` with a hardcoded 2s timeout and reports the timeout as "not installed". `codex.cmd` needs ~1.5s warm, so a loaded machine reports a working Codex as missing with no retry. Also tracked for agent-kit.


huntharo:
Superseded by #1751.

Closing this rather than iterating further. Two max-effort reviews on this branch found 30 confirmed defects, and the second one established that the compensating discovery module here does not do what it claims:

- the merge replace destroys the operator's configured `source` one line before the exemption added to protect it
- two renderer fixes cancelled each other, so the failure shape upstream actually emits was preserved by the merge and then discarded by the list filter
- the sibling rescue never reaches an already-spawnable command, so a native `codex.exe` install that misses upstream's 2s probe still reports Missing
- anchoring the version regex did not close the ranking hole it was written for, because the versions that drive ranking come from upstream's unanchored parser

A mutation matrix over the previous fix round is the clearest signal: of 16 fixes, **13 could be reverted with the whole suite still green.**

The root causes are upstream in `@pwrdrvr/agent-kit`, both now filed:

1. `buildPathCommandNames` returning `[command, ...PATHEXT]` on win32, so the scan stops on npm's sh shim — which `CreateProcess` cannot execute — and never reaches `codex.cmd`. Generic, so `git` and `gh` discovery have it too.
2. The hardcoded 2s `--version` probe timeout, plus a timeout being reported indistinguishably from "not installed".

Fix those and most of this branch deletes itself.

#1751 carries forward the parts that stand on their own and depend on neither: the `.ps1` → `.cmd` launch redirect at all four spawn sites, the chip-classification work that made the Codex panel readable, the restored editable path field, and the shared validated-candidate predicate.

Branch kept, not deleted — the withdrawn approach and the two review passes are worth having on record.

PR #1756, opened 2026-08-18
docs(desktop): record Token Miser code-mode gate transport findings

huntharo:
Stacked on #1733.

`PostToolUse` cannot replace a code-mode tool result — `PostToolUseFeedbackOutput::code_mode_result` returns `self.original`, so the gate's summary lands as a developer message while the parent model still reads the full original. The fix direction was a `PreToolUse` command rewrite. This PR records what verifying that design against `codex-cli 0.145.0` actually turned up, because two findings change the design.

**The rewrite is inseparable from auto-approval.** Controlled A/B/A: `updatedInput` alone is ignored, `permissionDecision: "ask"` plus `updatedInput` is ignored, and only `permissionDecision: "allow"` applies the rewrite. Wrapping a command therefore auto-approves it. This contradicts the checked-out Codex source and RTK's Claude Code hook, so it is a Codex-specific divergence.

**The existing bridge is unreachable from the sandbox.** Against Codex's real base Seatbelt policy, TCP loopback is `EPERM`. The Token Miser bridge is HTTP over `127.0.0.1`, so a rewritten in-sandbox command can only reach it under full access. `AF_UNIX` also fails by default but works once path-allowlisted through `network.allow_unix_sockets`; Linux's seccomp filter permits `AF_UNIX` already. macOS `sun_path` is 104 bytes, which the profile-state path is close to.

**Hook approval does not need the Codex CLI.** Trust lives at `[hooks.state."<key>"] trusted_hash`, and the Codex TUI writes it through the app server's `config/batchWrite` with `key_path: "hooks.state"`. PwrAgent already speaks that protocol.

The implication is that a deterministic wrapper spilling full output to a file needs no network, no sandbox hole, and no latency budget, with summarization moved out of band onto the retrieval tools that already exist. Finding 1 constrains every variant equally.

Docs only — no code changes. Open questions and reproduction notes are in the record.

### DIFF
--- a/apps/desktop/electron.vite.config.ts
+++ b/apps/desktop/electron.vite.config.ts
@@ -52,6 +52,10 @@ export default defineConfig(({ command }) => {
             "mcp-connection-bridge": resolve(
               __dirname,
               "src/main/mcp-connections/mcp-connection-bridge-entry.ts"
+            ),
+            "token-miser-hook": resolve(
+              __dirname,
+              "src/main/token-miser/token-miser-hook-entry.ts"
             )
           },
           output: {
--- a/apps/desktop/src/main/__tests__/backend-registry-token-miser.test.ts
+++ b/apps/desktop/src/main/__tests__/backend-registry-token-miser.test.ts
@@ -1,6 +1,11 @@
 import { mkdtempSync, rmSync } from "node:fs";
 import os from "node:os";
 import path from "node:path";
+import type {
+  AgentEvent,
+  NavigationSnapshot,
+  ThreadUsageLineRecord,
+} from "@pwragent/shared";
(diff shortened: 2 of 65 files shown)


huntharo:
Superseded by #1856.

This PR was opened against an earlier revision of `agent/token-miser`. That branch was rebased, stranding this PR on a base that no longer existed — which is why it shows 65 files and ~10k lines. Only one file was ever actually in it, `docs/plans/2026-08-18-token-miser-code-mode-gate.md`.

#1856 replays that file onto current `agent/token-miser`, updated to as-built: the design recorded here shipped on both sides, so the stale open items are folded into what landed and the Part 1 measurements are kept as the part that isn't recoverable from the code.

PR #1782, opened 2026-08-20
feat(desktop): give the Star Map chat card working mentions

huntharo:
Follow-up to #1780. That PR swapped the Star Map chat card's composer to the shared markdown editor and deliberately left mentions out — `CompactComposer` passed a module-level empty token array, so `$`, `@`, and `#` typed through as literal text.

This wires them up.

## What the card can honestly resolve

All three trigger characters, and every chip kind they produce in the full composer:

| Trigger | Source | Notes |
|---|---|---|
| `$skill` | `desktopApi.listSkills` via `useThreadSkills` | Thread-scoped — the skill list comes from the thread's linked directories, so two cards genuinely differ. Fetched on the first `$`, not on mount. |
| `@directory` | local navigation snapshot | Shared module-level cache: a dozen open cards make one fetch, not twelve. |
| `#thread` | navigation snapshot + `jumpSearchRemoteThreads` | Peer threads included — a card on a star map is usually open *because* of another instance. |
| `#123` | `thread.prs` off the same snapshot | Chip carries the PR's real status, same as the sidebar. |

**Left out: `@file` chips.** Not a scope call — `kind: "file"` was never a trigger-driven kind. In the full composer the `@` popover lists only tracked directories; file chips come from drag/drop, paste, and the native file dialog. So the card offers everything typing can reach.

## Structure

Two commits. The first is a pure extraction with no behaviour change — the mention plumbing that lived in `Composer.tsx` (12k lines) moves to modules a second surface can call:

- `composer-mention-tokens.ts` — token minting per chip kind, the index adjustment that keeps zero-width tokens aligned as the plain draft changes around them, and the markdown serializer. Verbatim move.
- `buildHashReferenceOptions` in `lib/hash-references.ts` — the local/peer merge and dedupe that was inline in Composer's `#` memo, now with its own unit tests.
- `HighlightedAutocompleteLabel.tsx` — its own module now that two surfaces render autocomplete rows.

The second adds `useComposerMentions` (which popover is open, which row is highlighted, where the caret lands) and `useComposerMentionSources` (the shared, lazy population). `CompactComposer` grew 17 lines and re-implements nothing.

`mentionSources` is **optional**. A host that supplies nothing keeps today's literal-text behaviour, so this is not a breaking change for future adopters — there's a test pinning that.

## Popover placement and the camera guards

The list renders **inside the card**, not portalled to `document.body`. That is what keeps it under `.star-map-chat-card` — the selector `isStarMapTypingTarget`, `shouldPanOnWheel`, and `shouldStartCanvasPan` all test for. A portalled list would sit outside all three and arrow keys or a wheel over an open popover would fly the camera. A test asserts all three guards directly against a rendered option.

No z-layer changes were needed and `star-map-z-layers.test.ts` is untouched: the popover uses the same in-card `position: absolute` / open-upward shape as the existing kebab menu, so it never leaves `.star-map-chat-card`'s stacking context. `.star-map-chat-card` clips its children, so the list is capped short enough to open over the transcript rather than out of the card.

## Behaviour notes

- Precedence and keyboard match the full composer: `$` → `@` → `#`, arrows clamp, Enter/Tab commit, Escape retires one popover, a trigger with no candidates yields to the next.
- Cold `#` anchors are carried over — `#` queries span spaces, so nothing else retires them and a `#` mid-sentence would otherwise keep the picker armed and the peer search re-firing for the rest of the line.
- Federated `#` search is **opt-in via `searchRemoteThreads`**, not the shared hook's implicit bridge fallback, so a source-less host can't fire a network search by typing `#`.
- A bounced send restores the draft *and its chips*, unless the operator has started something new.

## One test-infrastructure change

`Range.getClientRects` is stubbed in the renderer test setup. jsdom does not implement it, and ProseMirror asks a Range for its rects whenever `editor.commands.focus()` scrolls a selection into view — which the post-insert frame does. Synchronous tests never notice because teardown destroys the editor first; the moment a test awaits anything after inserting a chip, that frame fires while mounted and the missing method surfaces as an unhandled `TypeError` that fails the whole run rather than any one test. Empty rects are the honest answer: jsdom does no layout, so ProseMirror's scroll is correctly a no-op.

`ComposerTiptapInput.tsx` is unchanged, as expected.

## Tests

`CompactComposer.test.tsx` (+12) covers each trigger opening its picker, selection inserting a chip, the chip's markdown in the text handed to `onSend`, a no-match trigger staying literal, two chips at their own offsets in one draft, Escape handing Enter back to the send path, the ARIA wiring, lazy loading, and the source-less default. `StarMapChatCard.test.tsx` (+4) covers lazy `listSkills`, `@` reaching `startTurn` as markdown, the current thread never being offered by `#`, and the three gesture guards. `hash-references.test.ts` (+4) covers the extracted merge/dedupe.

One pre-existing test needed a harness fix: it captured a single `onAgentEvent` listener, and the card now has two subscribers, so the fake fans out instead.

## Verification

```
pnpm test apps/desktop/src/renderer   # 193 files, 3093 tests
pnpm typecheck
pnpm lint:eslint                      # 0 errors (44 pre-existing warnings)
pnpm lint:colors
pnpm lint:boundaries
```

All green, and the renderer suite was run repeatedly to confirm the run is stable.

**Not verified:** desktop E2E is not run on this machine, so nothing here was exercised in a real Electron window. Specifically unverified by automation: the popover's actual geometry inside the pan/zoom canvas at non-1.0 scale, and live federated `#` results from a real peer (the wiring is tested, the round trip is not).

### DIFF
--- a/apps/desktop/src/renderer/src/features/composer/CompactComposer.tsx
+++ b/apps/desktop/src/renderer/src/features/composer/CompactComposer.tsx
@@ -8,7 +8,10 @@ import {
 } from "react";
 import type { ThreadExecutionMode } from "@pwragent/shared";
 import { ComposerTiptapInput } from "./ComposerTiptapInput";
-import type { ComposerSkillToken } from "./ComposerInputTypes";
+import {
+  useComposerMentions,
+  type ComposerMentionSources,
+} from "./useComposerMentions";
 
 export type CompactComposerAction = {
   disabled?: boolean;
@@ -27,6 +30,12 @@ export type CompactComposerProps = {
   canSteer?: boolean;
   disabled?: boolean;
   executionMode?: ThreadExecutionMode;
+  /**
+   * Populations the `$` / `@` / `#` popovers pick from. Optional on
+   * purpose: a host that supplies nothing keeps the trigger characters as
+   * literal prose, so adopting this component never requires them.
+   */
+  mentionSources?: ComposerMentionSources;
   /** Thread's current model, shown as ambient state rather than a control. */
   model?: string;
   onInterrupt?: () => void;
@@ -48,13 +57,6 @@ const EXECUTION_MODE_LABELS: Record<ThreadExecutionMode, string> = {
 };
 
 /**
- * Module-level so the identity is stable: the Tiptap input re-syncs its
- * document whenever this prop changes identity, and a fresh `[]` on every
- * render would make that run on every keystroke.
- */
-const NO_SKILL_TOKENS: ComposerSkillToken[] = [];
-
-/**
  * Composer for surfaces with no vertical budget — currently the star map's
  * floating chat cards.
  *
@@ -73,9 +75,10 @@ const NO_SKILL_TOKENS: ComposerSkillToken[] = [];
  * rather than hydrating it on focus, which would cost a click and a caret
  * every time the operator moved between cards.
  *
- * Mentions are the one thing left out: `$skill`, `@file`, and `#thread`
- * need the backend list and directory index the card does not have, so no
- * skill tokens are ever supplied and the trigger characters stay literal.
+ * Mentions come from `useComposerMentions`, driven by whatever populations
+ * the host can honestly supply through `mentionSources`. Nothing about the
+ * pickers is re-implemented here: the triggers, ranking, token minting and
+ * markdown serialization are the same modules the full composer calls.
  *
  * Two moves buy back the height: secondary actions consolidate under a
  * single kebab, and model / reasoning effort render as right-aligned
@@ -83,7 +86,7 @@ const NO_SKILL_TOKENS: ComposerSkillToken[] = [];
  * state to be aware of, not controls to reach for.
  */
 export function CompactComposer(props: CompactComposerProps) {
-  const [draft, setDraft] = useState("");
+  const mentions = useComposerMentions({ sources: props.mentionSources });
   const [menuOpen, setMenuOpen] = useState(false);
   const menuRef = useRef<HTMLDivElement | null>(null);
   const { onSend } = props;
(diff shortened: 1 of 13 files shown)


huntharo:
## Performance check

Measured rather than assumed, since the composer sits on a surface that re-renders while a turn streams.

**The multiplier.** `CompactComposer` re-renders on *every* streamed delta — measured `rendersPerDelta: 1` against `item/agentMessage/delta`. At Codex's ~444 deltas/sec that is ~444 composer renders/sec per streaming card, so anything added to the render path runs at that rate.

**What this PR adds per render**, timed over 2000 iterations on a realistic 4-line draft:

| Work | Cost |
|---|---|
| `inputRef.current.selectionStart` (walks the ProseMirror doc) | 1.08 µs |
| The three trigger regexes (`$`, `@`, `#`) | 1.01 µs |
| **Total added per render** | **~2.1 µs** |

At 444 Hz that is **~0.09% of one core** per streaming card. A full composer re-render costs hundreds of µs in the same environment, so the mention work is well under 1% of a render that already happens. Candidate lists are memoized on the query string, so narrowing a search does not re-rank on unrelated renders, and `serializeDraftWithSkillTokens` early-returns the draft unchanged when there are no chips.

**Per-event cost.** The card now has two `onAgentEvent` subscribers instead of one (`useThreadSkills` added the second). Its filter is `backend !==`, `federationTargetsEqual` (three property compares), then two method-string compares before returning — roughly five comparisons per event per card. Negligible even at 12 cards × 444 events/sec.

**IPC.** Both loads are lazy and neither happens on mount: `listSkills` fires on the first `$`, and the navigation snapshot on the first `@`/`#`, shared module-wide so a dozen open cards make one fetch rather than twelve. For reference, the app's own background navigation refresh is every 5 minutes; this adds at most one fetch per 10s and only while a picker is actually being opened. No SQLite writes are added, so no write-budget movement.

**One thing worth calling out, now pinned by a test.** The cache's staleness guard turned out to be load-bearing for correctness, not just for caching: a host builds its `ComposerMentionSources` from the population the hook returns, so a completed load changes that object's identity and re-runs the effect that asked for the load. Unguarded that is a fetch loop running at typing speed — and it would show up as IPC churn, not as an error. Measured at exactly one fetch across a whole `@` session, and `settles a whole picker session on one snapshot fetch` now pins it.

**Caveat on the numbers.** These were taken in jsdom, so absolute render times are much slower than Electron. The 2.1 µs figure is pure JS (a ProseMirror doc traversal and three regexes, no layout), so it should carry over closely; the full-render denominator would only get larger in a real browser's favour.

PR #1876, opened 2026-08-31
fix(messaging): ignore Discord messages addressed elsewhere

huntharo:
## Summary

- I prevent PwrAgent from starting a turn for a shared Discord message that begins with an exact user-ID mention of someone else, even when the conversation accepts every ambient message.
- I carry the provider-neutral `addressedToOtherParticipant` signal from the Discord adapter into shared-conversation routing and apply it before pending replies or turn admission.
- I added adapter and controller regressions using the incident's foreign Discord user ID.

## Verification

- `pnpm test packages/messaging/providers/discord/src/__tests__/discord-adapter.test.ts apps/desktop/src/main/__tests__/messaging-controller.test.ts` (455 passed)
- `pnpm typecheck`
- `pnpm lint:eslint` (no errors; existing unrelated warnings remain)
- `pnpm lint:boundaries`

## Notes

- I confirmed the current adapter already matches the configured bot ID exactly; it was not doing a display-name prefix match. The bug was that a foreign mention remained ordinary ambient text, which `every_message` mode then admitted.
- Direct messages retain their existing behavior. The new suppression applies only to shared conversations that a provider positively identifies as addressed to another participant.

### DIFF
--- a/apps/desktop/src/main/__tests__/messaging-controller.test.ts
+++ b/apps/desktop/src/main/__tests__/messaging-controller.test.ts
@@ -7610,6 +7610,32 @@ describe("MessagingController", () => {
     );
   });
 
+  it("ignores shared messages explicitly addressed to another participant", async () => {
+    const harness = await createHarness({
+      responseModeForConversation: () => "every_message",
+    });
+    const channel = buildTopicChannel("500");
+    await harness.store.upsertBinding({
+      id: "binding:telegram:topic:-1001:500:codex:thread-1",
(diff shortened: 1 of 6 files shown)


huntharo:
Closing as superseded by the Discord response-mode work in #1875/#1874. The incident was caused by Discord lacking the configurable mention-only inheritance hierarchy, not by display-name matching. The feature thread is now auditing global, guild/server, channel/native-thread, and bound-thread overrides.

PR #1877, opened 2026-08-31
feat(desktop): add per-review location choice

huntharo:
## Summary

- add an explicit Review location choice for inline parent reviews or PwrAgent-managed child reviews
- force and explain managed-child routing for cross-provider, ACP, and secondary-workspace reviews while gating unsupported reviewers
- propagate the explicit run mode through IPC, queued, scheduled, messaging, and Star Map review starts
- retire the global experimental managed-review switch while preserving config round-trip compatibility

## Validation

- 1,650 affected tests passed across 12 complete test files
- pnpm typecheck
- pnpm lint:eslint (0 errors; existing warnings only)
- pnpm lint:boundaries
- pnpm lint:colors
- pnpm build

## Stack

Base: #1869 (fix/inline-review-parent-turn)


huntharo:
superseded

PR #1881, opened 2026-08-31
feat(desktop): transfer cross-thread attachments

huntharo:
## Summary
- stage image and file inputs in PwrAgent-owned content-addressed storage
- transfer attachment bytes in authenticated binary federation frames without base64 JSON payloads
- propagate source-turn attachments through send, steer, handoff, create-instance, and Star Map intake paths
- reject peer-controlled paths and enforce chunk geometry, memory limits, size limits, and SHA-256 validation

## Validation
- desktop TypeScript typecheck
- focused staging, transport, bridge, agent-tool, send-message, and handoff tests
- federation router tests
- ESLint typed and untyped checks
- dependency-boundary lint
- Codex-storage boundary lint
- git diff check


huntharo:
Superseded by #1882, which includes this transport commit together with the Star Map intake UI and renderer-to-main staging boundary.

PR #1903, opened 2026-09-01
fix(desktop): keep quit blocker titles stable

huntharo:
## Summary
- retain a known quit-blocker title when an authoritative refresh temporarily omits optional title data
- continue accepting blocker removals, count changes, and later non-empty title updates
- add a regression test reproducing the title-to-UUID downgrade caused by alternating title-resolution outcomes

## Root cause
The durable queue polls every 500 ms. Missing labels are resolved through a best-effort listThreads call bounded by a 1.5-second timeout, so cache hits can return titles while slower refreshes fall back to thread IDs. Replacing the whole renderer snapshot made the visible label oscillate. This fix keeps membership authoritative while preventing optional presentation data from degrading.

## Validation
- pnpm test apps/desktop/src/renderer/src/features/notifications
- pnpm --filter @pwragent/desktop typecheck
- targeted ESLint
- pnpm lint:boundaries

### DIFF
--- a/apps/desktop/src/renderer/src/features/notifications/QuitBlockerQueueToast.tsx
+++ b/apps/desktop/src/renderer/src/features/notifications/QuitBlockerQueueToast.tsx
@@ -20,6 +20,27 @@ const GROUPS: ReadonlyArray<{
   { kind: "action", heading: "Environment actions" },
 ];
 
+function retainKnownTitles(
+  current: QuitBlockerQueueSnapshot | undefined,
+  next: QuitBlockerQueueSnapshot,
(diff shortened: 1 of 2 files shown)


huntharo:
Superseded by #1907, which now carries this commit rebased onto main.

The stack collapsed because #1907 grew a dependency on `getCachedThreadSummary`, which landed in main with #1896 and is not on this branch. Rather than restack, the two PRs are now one: `f487978ca fix(desktop): keep quit blocker titles stable` is the first commit of #1907, unchanged apart from the rebase.

Reopen this if you would rather review the renderer-side retention separately.

PR #1907, opened 2026-09-01
refactor(desktop): give thread information its own store

huntharo:
## Summary

The quit card's titles blinking between name and UUID every ~3s was the symptom.
The cause was that this product had no place to keep what it knows about a
thread — thread information lived inside the caches that answer *queries*, and
those are correctly discarded on every mutation.

This PR gives thread information its own store and moves the callers onto it.

- **`ThreadInfoStore`** (`app-server/thread-info-store.ts`) holds one entry per
  thread, with a per-field observation sequence and source. Identity is
  `instanceId::backend:threadId`.
- **Positive facts only.** An omitted field is not a deletion, an empty title is
  not news, and a `fallback` source means the title *is* the UUID — recording it
  would overwrite a real name with nothing. Title and source are validated
  together so a fallback cannot arrive alone and downgrade a held name.
- **Sequences are reserved before an async read begins**, so a listing overtaken
  by a rename cannot revert it on late completion. Monotonic counter, never
  `Date.now()`.
- **Callers declare freshness.** `findThreadForWorkspaceHandoff` takes
  `freshness: "provider" | "last-known"` — am I about to act on this, or only
  display it?
- **Archival is an observation**, recorded from the notification at the event
  fan-out, not rediscovered by invalidating and re-listing.
- **Peer names get the same treatment**, keyed by instance so two peers reusing
  a thread id cannot answer for each other.

Only the thread or the peer going away removes an entry. Cache invalidation
never does.

Stacked on #1903.

## What was actually wrong

Four defects, established by measurement rather than assumption:

1. **Invalidation was amnesia.** ~46 sites clear the list cache; every one was
   also erasing the only record of what threads were called.
2. **Point queries were answered by full-collection walks.** 16 sites wanted one
   row and rebuilt every row to get it.
3. **Query-shaped keys fragmented identical reads.** 6 concurrent callers asking
   the same question produced more than one provider call.
4. **Absence was conflated with unknown.** `title: undefined` meant both "has no
   title" and "we knew it and this lookup didn't return it".

The headline case: `turn/started` invalidates the list cache, and
`active-turn-branch-adoption` then drove a complete paged provider `thread/list`
to read a single field — **one full listing per turn boundary.** Quantified by
reverting the fix and re-recording the budgets:

```
turn-lifecycle-label-reads        5 -> 0
ten-turns-warm-thread            10 -> 0
worktree-turns-after-enrichment   5 -> 0
```

## Provider-read budgets

Provider reads are now budgeted the way SQLite writes already are here: a
checked-in JSON of exact counts per named scenario
(`__tests__/fixtures/thread-read-budgets.json`), asserted by
`expectThreadReadBudget`, failing on deviation **in either direction** — a count
that drops is as much a change to review as one that rises. Re-record with
`UPDATE_THREAD_READ_BUDGETS=1`.

| Scenario | Lists | Enrichments |
|---|---|---|
| Ten complete turns on an already-observed thread | 0 | 0 |
| Five turns with label reads between every lifecycle event | 0 | 0 |
| Five turns on an already-enriched worktree thread | 0 | 0 |
| Twenty quit-dialog polls against an in-progress turn | 0 | 0 |
| Fifty synchronous label lookups for a thread observed once | 0 | 0 |
| A second window's navigation refresh against a warm cache | 0 | 0 |
| Six concurrent callers asking for the same unenriched listing | 1 | 0 |
| Eight terminal notifications across two never-listed threads | 1 | 0 |
| One navigation refresh over a single worktree-backed thread | 1 | 1 |

The three nonzero rows are the intended shape. The worktree scenarios exist
because no other fixture reaches directory enrichment, and a budget counter that
cannot move proves nothing.

## Tests

New: `thread-info-store.test.ts` (store contract),
`thread-read-budget.test.ts` (the budgets above), plus remote ordering and
isolation cases in `remote-thread-summary-cache.test.ts` and a reservation-order
assertion at the navigation-snapshot IPC seam.

**Every new behavioral test was falsified before being kept** — the fix was
reverted, the test watched to fail, and the fix restored. A test that passes
against the defect it names is not coverage.

## Validation

- `pnpm test` — **exit 0**, 644 files, 9394 passed, 6 skipped
- `pnpm typecheck` — exit 0
- `pnpm lint:eslint` — exit 0
- `pnpm lint:boundaries` — exit 0
- `pnpm lint:codex-storage` — exit 0
- `pnpm lint:colors` — exit 0

No new desktop E2E: this adds no renderer, dialog, IPC-shape, or interaction
contract — it changes where the main process keeps what it already knew. PR
#1903's renderer-side monotonic retention stays in place as a second
presentation-layer defense.

## Notes for review

- No new persistent state. Both stores are process-local, so there are no new
  SQLite writes or write budgets.
- `docs/plans/2026-08-31-quit-blocker-title-read-model.md` was rewritten in this
  branch: it opened with "This is a scoped correction to the existing quit path",
  which stopped being true. It is unmerged and was added by this PR's own first
  commit.

### DIFF
--- a/apps/desktop/src/main/__tests__/app-server-ipc.test.ts
+++ b/apps/desktop/src/main/__tests__/app-server-ipc.test.ts
@@ -96,6 +96,7 @@ const federationMock = vi.hoisted(() => {
     })),
     getWorktreeUnpublishedCommitDiff: vi.fn(async () => ({})),
   };
+  let nextThreadNameObservation = 0;
   const remoteThreadSummaries = {
     resolvePinnedThreads: vi.fn(
       async (): Promise<{
@@ -111,6 +112,7 @@ const federationMock = vi.hoisted(() => {
     searchForJump: vi.fn(async () => ({ results: [] })),
     threadFromPeer: vi.fn(async (): Promise<unknown> => undefined),
     rememberThreadNames: vi.fn(),
+    reserveThreadNameObservation: vi.fn(() => (nextThreadNameObservation += 1)),
     invalidate: vi.fn(),
   };
   return {
@@ -1968,6 +1970,11 @@ describe("app server ipc", () => {
         executionMode: "default" as const,
       },
     };
+    // These three carry calls over from earlier cases, and the assertions below
+    // read call zero rather than any call.
+    federationMock.runtime.remoteNavigationSnapshot.mockClear();
+    federationMock.remoteThreadSummaries.rememberThreadNames.mockClear();
+    federationMock.remoteThreadSummaries.reserveThreadNameObservation.mockClear();
     federationMock.runtime.remoteNavigationSnapshot
       .mockResolvedValueOnce(snapshot)
       .mockResolvedValueOnce(snapshot);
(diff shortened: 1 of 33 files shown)


huntharo:
Superseded by #1916, which is the same work rebased onto main as a single PR against main rather than a stack on #1903.

Collapsing was necessary: this branch grew a dependency on `getCachedThreadSummary`, which landed in main with #1896 and was not on #1903's branch. GitHub declines to retarget a stacked PR from either the REST or GraphQL API, so #1916 is a fresh PR on a new branch with identical commits.

PR #1928, opened 2026-09-01
fix(desktop): number nested ordered lists as 1.2 / 2.2.3

huntharo:
## Summary

- Nested ordered lists no longer restart at `1` at each indent.
- Transcript and composer `ol` markers now use CSS `counters()`, so the list in the report reads `1`, `1.1`, `1.2`, `2`, `2.1`, `2.2`, `2.2.1`, `2.2.2`, `2.2.3`.
- `role="list"` stays on transcript ordered lists so Safari does not drop list semantics when `list-style` is none.

Follow-up to #1893, which preserved the nested structure but still showed a fresh `1.` at every level.

## Verification

- `pnpm test apps/desktop/src/renderer/src/styles/__tests__/theme-contract.test.tsx apps/desktop/src/renderer/src/features/thread-detail/__tests__/thread-markdown.test.tsx` (124 passed)

### DIFF
--- a/apps/desktop/src/renderer/src/features/thread-detail/ThreadMarkdown.tsx
+++ b/apps/desktop/src/renderer/src/features/thread-detail/ThreadMarkdown.tsx
@@ -595,7 +595,11 @@ export const ThreadMarkdown = memo(function ThreadMarkdown(props: ThreadMarkdown
         );
       },
       ol(listProps) {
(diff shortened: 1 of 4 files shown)


huntharo:
Closing this — the nested list in the thread was a renderer validation of #1893, not a request for outline numbering (1.2 / 2.2.3). Sorry for the misfire.

PR #1968, opened 2026-09-03
fix(desktop): hold queued turns after provider failures

huntharo:
## Summary

- I made failed-turn recovery provider-neutral: queued user messages are held after a Codex or ACP turn failure and require an explicit Retry instead of being submitted into a potentially ongoing outage.
- I made stale steering fallbacks durable and visibly held, added local and federated queue-release APIs, and preserved FIFO order across renderer restarts and remote snapshots.
- I tightened ACP terminal handling so a provider that reports `turn_finished` or `turn_completed` but leaves `session/prompt` open fails cleanly after a short grace period, records the failure, publishes idle state, and surfaces the existing sticky backend-error notice.
- I fixed ACP active-turn lookup so a steer can reach a provider while its provisional turn ID is still pending. This is the state the affected Grok thread was in.

## Verification

- `pnpm typecheck`
- `pnpm lint:eslint` (passes with 45 existing warnings and no errors)
- `pnpm lint:sql`
- `pnpm lint:codex-storage`
- `pnpm lint:colors`
- `pnpm lint:boundaries`
- `pnpm licenses:check`
- Focused provider/queue/scheduler/IPC/federation/renderer suite: 11 files, 492 tests passed with `--maxWorkers=1`
- Full backend-registry suite: 613 tests passed with `--maxWorkers=1`

## Notes

- The affected thread did use the custom Grok build. The hour-long state was caused by PwrAgent waiting on an ACP `session/prompt` request after Grok had already emitted a terminal update; the old timeout was one hour.
- A full local macOS `desktop-main` run reaches unrelated passing updater/SQLite tests and then the Vitest process exits with native status 139 without reporting a failing test. The focused affected suites and the complete backend-registry suite pass; CI is the clean-environment full-suite signal for this PR.


huntharo:
Addressed all review findings in signed commit `ecce7ce49`:

- persist live `held` queue events into the durable scheduled-action row, including `manualReleaseRequired`
- retry the existing deterministic queue entry and keep blocked/busy retries visibly held
- place stale-steer holds at the composer head and preserve that placement during reconciliation
- allow Delete/Edit to cancel a durable hold after its process-local queue entry is lost on restart
- add exact SQLite budgets for both the failure-boundary hold and an explicit successful Retry

Write budgets:

- queued action → held: 1 commit, 1 statement, 1 row, ~16 KB observed WAL
- explicit held Retry → started: 3 commits, 3 statements, 3 rows, ~45 KB observed WAL
- zero timer or idle writes

Validation:

- 166 main provider/queue/scheduler/IPC/federation tests passed
- 331 composer/projection/notification tests passed
- all 40 SQLite write-budget tests passed
- workspace typecheck, ESLint, SQL/storage/color/boundary lints, and license checks passed

PR #2001, opened 2026-09-06
fix(desktop): bound federation collections and progressive browse

huntharo:
## Summary

- I replaced full Federation navigation snapshots with distinct bounded query pages, authoritative counts and placement, and independent exact detail, FIFO and scheduled projections. Main and remote sidebars, Star Map, search, mentions, project tools and exact thread resolution use the new contracts.
- I preserved off-page selection/history, owner-ranked Attention, local drafts, background accepted queue release and relative pin ordering. Local and remote consumers share read admission, cancellation and retained-backing budgets; unchanged retained ranges avoid retransferring displayed rows.
- I retired remote full-list/snapshot fallbacks and single-frame peer directories. Unsupported alpha participants get actionable upgrade errors. All participating viewers, owners and gateways must run this cutover.

## Verification

- Earlier checkpoint `8be8c1979` passed the full unit suite: 754 files, 10,804 passing tests and 7 skipped. Workspace typecheck, full ESLint, dependency boundaries, SQL, storage, colors, licenses and Electron-version checks passed at that checkpoint. Current repair validation is recorded below.
- I ran 13 Electron E2E cases covering remote windows, Star Map flow and large history, transcript readiness, and queued-review release; all passed.
- I tested the visible apps on the host with isolated `dev` and `work` profiles. Paging returned ten rows, an unchanged refresh transferred only 681 bytes, and continuation had no duplicates. Search, both Star Map layouts, a remote window, and disconnect/reconnect passed. I left the default profile and its messaging lease untouched and closed the test apps cleanly.
- I measured zero SQLite commits for read-only navigation. A separate 10,000-thread allocation probe reduced the largest serialization from 5.25 MB to 53 KB. Serialized-backing budgets and sampled heap measurements are reported separately.

## Notes

- This draft failed live operator acceptance and is not ready for merge. The real M4/M5 Federation exposed missing cross-owner children, duplicate project folders, unstable pins/pages, query-pool pressure and oversized selected detail blocking Send. Repairs require validation against the same data. GitHub CI completion and explicit operator approval remain merge gates; I have not merged it.
- [Operator acceptance checklist](docs/federation-navigation-acceptance.md), [protocol contract](docs/federation-navigation-v2.md), and [measured budgets](docs/federation-collection-budgets.md).
- Initial selected history may still be multi-megabyte; provider pagination and large content are preserved. Arrangement tombstones remain lossless, and peer-directory broadcast work remains O(peers²).
[...]

PR #2030, opened 2026-09-08
feat(desktop): capture directory enrichment probe diagnostics

huntharo:
Directory-enrichment Git spawns dominate the reported CPU slice, but the profile cannot distinguish cold paths, cache invalidations, failed observations, or recreated cache owners. The existing pending map already coalesces in-flight requests.

Add bounded in-memory accounting for enrichment callers, cache decisions, cache-owner IDs, publication outcomes, and Git command starts, successes, failures, and elapsed time. Main-process hot CPU captures write and register a sibling `main-hot-NNNN.diagnostics.json`; failure to collect auxiliary diagnostics does not discard the CPU profile.

Retain 60 two-second buckets, with 32 directory/caller/reason/owner rows per bucket and explicit overflow totals. The documentation explains retention, attribution limits, bucket boundaries, and elapsed-time interpretation. Collection adds no timers, per-call logging, Git commands, or SQLite writes (0 MB/day). Probe behavior and cache invalidation remain unchanged.

Validation: 885 tests passed across seven directory, client, registry, and CPU-capture suites. Full `pnpm lint` passed, including typechecks and dependency boundaries; ESLint reported 0 errors and 47 existing warnings. `git diff --check` passed. The running application was not restarted.

Windows CI exposed an existing cleanup race in the Electron-environment detached-action test: complete output appeared before the command owner released its cwd. The test now waits for `onDetachedExit` before checking output and deleting the fixture directory. Both affected suites pass locally (66 tests), along with ESLint and typechecks. This scoped repair does not change process launch behavior or increase retries/timeouts; the separate Git-for-Windows startup failures remain subject to Windows CI verification.

### DIFF
--- a/apps/desktop/src/main/__tests__/backend-registry.test.ts
+++ b/apps/desktop/src/main/__tests__/backend-registry.test.ts
@@ -14394,6 +14394,7 @@ script = "echo setup"
     });
 
     expect(enrichThreadDirectories).toHaveBeenCalledTimes(1);
+    expect(enrichThreadDirectories).toHaveBeenCalledWith(expect.any(Array), "missing-worktree-backfill");
     expect(
       overlaysByThreadId["project-a-worktree-2"]?.extraLinkedDirectories[0],
     ).toMatchObject({
@@ -15142,7 +15143,7 @@ script = "echo setup"
     });
 
     expect(enrichThreadDirectories).toHaveBeenCalledTimes(1);
-    expect(enrichThreadDirectories).toHaveBeenCalledWith([cheapThread]);
+    expect(enrichThreadDirectories).toHaveBeenCalledWith([cheapThread], "selected-thread");
     await expect(
       overlayStore.getThreadOverlayState({ backend: "codex", threadId: "thread-1" }),
     ).resolves.toMatchObject({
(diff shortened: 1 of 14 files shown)


huntharo:
Operator-requested coordination with #2001: a new real-data CPU capture identifies a separate navigation overlay stall. At 2026-09-08T01:16:44.644Z the process measured 72.62% CPU over 2.001s. Weighted samples show `buildLocalNavigationQueryIndex -> readNavigationQueryIndex` at 1,144.9 ms inclusive, including `listManagedSubAgentThreadKeys` at 773.1 ms and `getBackend` at 167 ms. The retained 75.163-second recording was otherwise mostly idle (73.354 seconds).

This is **not established as a #2001 regression**: main already invokes the helper from `reconcileNavigationSnapshot` and scans overlay payloads for managed monitor/subagent identities. A Python probe of only the first query returned about 15.18 MB on main versus 447 KB with #2001's compact projection; it does not measure full Electron runtime cost or prove a regression.

A child based on current `origin/main` now owns the inherited overlay scan/materialization fix and a separate PR. #2001 owns investigation of query frequency/admission. #2027 remote/SSH deduplication and this PR's directory-enrichment diagnostics appear orthogonal. Please flag any existing work on `SqliteOverlayStore.listManagedSubAgentThreadKeys` / `getBackend` to avoid duplicate implementation. Direct thread messages and steers were rejected by the active-writer guard, so recording coordination here.

PR #2136, opened 2026-09-14
ci(tests): require the macOS E2E lanes in the Desktop E2E gate

huntharo:
## Problem

`desktop-e2e-result` carries `name: Desktop E2E`, which is the required status-check context in both the **Protect main** and **Protect maintenance release branches** rulesets. Its one step asserted only `needs.desktop-e2e.result`.

So the four `macOS Desktop E2E (lane N of 4)` jobs ran on every code-impacting PR — burning self-hosted Tart capacity — while **nothing required them**. A macOS-only regression went green on the gate. That is worse than having no macOS lane, because the gate looks like coverage.

## Change

`desktop-e2e-macos` joins `needs`, and the step asserts its result alongside the Linux one. Both platforms are evaluated before the step exits, so a red gate says which platform broke rather than stopping at the first check and hiding whether the other was healthy.

The two properties the job's header comment calls out are preserved:

- **`name: Desktop E2E` is byte-identical.** Renaming it would silently detach the ruleset rule.
- **`if: always() && ...` is unchanged.** `always()` is what stops a failed lane turning the aggregate into a skip — and a skipped required check is treated as satisfied, so without it the gate fails open exactly when it matters. That reasoning is unchanged by the second matrix: `always()` is what makes the job run at all once *any* `needs` entry is unsuccessful, and each entry is then read from `needs.*.result` inside the step rather than gating entry to the job. The comment has been updated to say so.

## The fork guard, which needed handling

`desktop-e2e-macos` carries a guard the Linux lane does not:

```yaml
github.event_name == 'push' ||
(github.event_name == 'pull_request' &&
 github.event.pull_request.head.repo.full_name == github.repository)
```

A PR from a public fork must not execute arbitrary code on a self-hosted machine, so the job skips there and reports `skipped` — for a reason that is not a failure. This repo is public and MIT, so that path is live.

- A bare `== 'success'` would have made `Desktop E2E` **permanently red for every fork contributor**.
- Accepting `skipped` outright would instead swallow the `skipped` that a *failed* `macos-install-deps` produces — a real failure, silently passed.

The step therefore mirrors the guard into `MACOS_E2E_EXPECTED` and requires `success` only when the lane was expected to run; when it was guarded off, only `skipped` is accepted. GitHub Actions has no YAML anchors, so that condition is necessarily a copy of the one on `desktop-e2e-macos` — the comment says to change both.

## Windows stays out, deliberately

`windows-e2e` is **not** added. It landed today (#2112) as, in its own words, "an **advisory** Windows Desktop E2E lane." Gating merges on a lane that is one day old and explicitly advisory would strand PRs on its flakes rather than on real regressions. The aggregate now records why it is absent and what would justify promoting it, so the omission does not read as an oversight. (For contrast, PwrSnap's Windows E2E has existed since 2026-06-08.)

## Verification

No macOS lane was deliberately broken to watch this go red — by request. What was verified:

**The gate logic, against a truth table.** The step body was run under `bash` across 11 combinations. `needs.<job>.result` for a matrix job is `success` only when every lane succeeded, and `failure` if any lane failed.

| `desktop-e2e` | `desktop-e2e-macos` | expected to run | exit | note |
|---|---|---|---|---|
| success | success | true | **0** | the green path |
| success | **failure** | true | **1** | ← the case this PR exists for |
| failure | success | true | 1 | |
| failure | failure | true | 1 | both errors reported, not just the first |
| success | skipped | true | 1 | `macos-install-deps` failed |
| success | cancelled | true | 1 | |
| skipped | success | true | 1 | `install-deps` failed |
| success | skipped | false | **0** | fork PR — stays mergeable |
| failure | skipped | false | 1 | |
| success | failure/success | false | 1 | guard and reality disagree |

**The workflow itself.** Parsed with `js-yaml`: `name` is exactly `Desktop E2E`; `needs` is `[changes, desktop-e2e, desktop-e2e-macos]`; the folded `MACOS_E2E_EXPECTED` collapses to a single line with no embedded newline (more-indented continuation lines in a `>-` scalar would have preserved one); its expression body is character-identical to the guard on `desktop-e2e-macos`; the `needs` graph is acyclic with no unknown job names.

**Why the failure path fires.** It is the same mechanism already proven in production by the Linux half of this same step, which has been gating merges from this job since it was introduced. The only new input is a second `needs.*.result`, read identically.

## Branch protection

Checked, since a per-lane job name required *in addition* to the aggregate would be a repository setting and not something to fix here.

`main` has **no classic branch protection**; it is governed by the **Protect main** ruleset (id `16543120`). It and **Protect maintenance release branches** (id `20676893`) each require exactly:

> `Lint`, `Build`, `Test`, `Desktop E2E`

**No per-lane job name is required in either ruleset**, so nothing needs a human here — the aggregate is the sole E2E gate, and widening it widens the gate.


huntharo:
## This PR's own CI is evidence for both halves of the decision

**The gate now waits on macOS, and it discriminates.** `Desktop E2E` ran for 39s *after* all four `macOS Desktop E2E (lane N of 4)` jobs finished — it could not have done that before this change, since it did not depend on them. All four macOS lanes and both Linux lanes passed, and the aggregate passed.

**And `Windows Desktop E2E (lane 3 of 4)` failed on this run — while `Desktop E2E` correctly stayed green.** That is the deliberate Windows exclusion working, and it is a live demonstration of why it is deliberate.

This PR changes **only `.github/workflows/ci.yml`**. It cannot have broken a Windows Playwright spec. The failure is in `star-map-composer-attachments.spec.ts`:

```
Error: Composer refused the keystroke: block=detail:none mount=1->1 identity=[null,"codex","threa…
Error: locator.fill: Element is not an <input>, <textarea> or [contenteditable] element
```

That is a renderer/composer readiness race — the "optimistic renderer synchronization" Windows-only class `CLAUDE.md` calls out by name — not a regression in this diff.

**Had `windows-e2e` been added to the aggregate, this workflow-only PR would be blocked right now by an unrelated flake.** That is precisely the "strand PRs on flakes" outcome the exclusion is there to prevent, observed on day one of the lane's existence. The lane stays advisory until it has a green baseline worth trusting.

The flake itself is real and worth fixing on its own terms — as a readiness/ownership fix, not by retries or longer timeouts — but it is out of scope here.


---

### Addendum: job timings prove the new dependency is live

"Everything was green" would not by itself show the aggregate actually *waits* on macOS. The scheduling does — GitHub does not create a dependent job until its `needs` resolve:

| Job | started | completed |
|---|---|---|
| Linux Desktop E2E (lane 1 of 2) | 15:40:20 | **15:45:58** ← last Linux lane |
| macOS Desktop E2E (lane 1 of 4) | 15:40:55 | 15:43:12 |
| macOS Desktop E2E (lane 2 of 4) | 15:44:45 | 15:46:54 |
| macOS Desktop E2E (lane 4 of 4) | 15:45:07 | 15:48:27 |
| macOS Desktop E2E (lane 3 of 4) | 15:45:58 | **15:48:34** ← last macOS lane |
| **Desktop E2E (aggregate)** | **15:48:37** | 15:49:16 |

The aggregate started **3 seconds after the last macOS lane finished**, and **2m39s after the last Linux lane had already finished**. On `main` today — where it depends on Linux alone — it would have started at ~15:46:00 and reported `success` while the macOS lanes were still running.

That is the structural change: the required check's conclusion is now a function of the macOS lanes' results. A macOS `failure` reaches the step as `needs.desktop-e2e-macos.result`, and the step exits 1 on any value other than `success` when the lane was expected to run.

huntharo:
## Updated: gate switched to the fail-closed form

Adopted from pwrdrvr/PwrSnap#599, which solved this independently and better.

The previous commit mirrored `desktop-e2e-macos`'s whole scheduling expression into `MACOS_E2E_EXPECTED`. That had a fail-open I missed: it enumerated the events on which macOS *should* run, so any event neither job anticipated — a `merge_group` or `workflow_dispatch` trigger added later — made it `false`, and the gate would then have accepted **any** macOS result. macOS would have dropped out of the required check silently. That is the same "looks like coverage" failure this PR exists to fix, reintroduced one layer up.

Now it tests only for the fork PR and requires success otherwise:

| Linux | macOS | fork PR | exit | |
|---|---|---|---|---|
| success | success | false | **0** | green path |
| success | failure | false | 1 | |
| success | **skipped** | **false** | **1** | ← the fail-open, now closed |
| success | skipped | true | **0** | fork PR, with a `::warning::` that macOS is unverified |
| success | failure | true | 1 | |
| failure | * | * | 1 | |

Verified across 9 cases. `Desktop E2E` passes on this run with all four macOS lanes green.

## Windows, two runs in

This PR changes **only `.github/workflows/ci.yml`**, and `Windows Desktop E2E (lane 3 of 4)` has now failed on **both** of its runs — with **different** failures each time:

- Run 1 — `star-map-composer-attachments.spec.ts`: `Composer refused the keystroke: block=detail:none mount=1->1`, then `locator.fill: Element is not an <input>, <textarea> or [contenteditable]`.
- Run 2 — `review-output-rendering.spec.ts:94`: `expect(getSelection().toString()).toContain("Preserve async pasted images")` received `"P2"` after a 5s predicate timeout.

Two different specs failing across two runs of a diff that cannot affect either one. That is not a single flaky assertion — shard 3 carries several specs that are unstable on Windows, and both failures are readiness/timing shaped: content asserted before the renderer has settled.

Had `windows-e2e` been added to this aggregate, this workflow-only PR would have been blocked twice, for reasons unrelated to it. The advisory exclusion is doing exactly what it was put there for. The lane joins the gate when it has a green baseline worth trusting — tracked in the comment on `desktop-e2e-result`.

Both failures deserve an ownership/readiness fix rather than retries or longer timeouts, per the Windows guidance in `CLAUDE.md`. Out of scope here.

huntharo:
## Rebased onto `main`, and brought to parity with the merged sibling PRs

**Rebase:** clean. `main` had moved 16 commits, **none** of them touching `.github/workflows/ci.yml`, so there were no textual or logical conflicts to resolve.

**Parity:** pwrdrvr/PwrGit#266 and pwrdrvr/PwrSnap#599 have both merged, and review on the former improved the aggregate in ways this repo's copy never picked up. Adopted here so the three cannot drift:

- `set -euo pipefail` — a future env-var rename becomes a named error instead of an empty comparison that fails closed with a blank message.
- `say`/`require` helpers instead of copy-paste per-platform blocks.
- Results mirrored into `$GITHUB_STEP_SUMMARY`, so the checks page names the platform that broke — and shows the healthy ones and the fork-PR state — without opening a job log.

The gate script is now **byte-identical to PwrGit's apart from its Windows `require`** (verified by diffing the two parsed `run` blocks). Promoting Windows here is that one call plus one `needs` entry.

Re-verified across 8 cases after the change, including the fail-closed path this PR's previous commit added:

| Linux | macOS | fork PR | exit | |
|---|---|---|---|---|
| success | success | false | **0** | green path |
| success | failure | false | 1 | |
| success | **skipped** | **false** | **1** | fails closed on an unanticipated event |
| failure | success | false | 1 | |
| failure | failure | false | 1 | both reported, not just the first |
| success | skipped | true | **0** | fork PR, with a `::warning::` that macOS is unverified |
| success | failure | true | 1 | |
| failure | skipped | true | 1 | |

## Windows still stays out — now with a measurement behind it

I checked whether the lane had become stable enough to gate. It has not.

Across the **20 `main` CI runs** after it landed — 80 real lane executions, no skips:

| | |
|---|---|
| Lane pass rate | 72/80 = 90% |
| **Run-level** pass rate (what the gate sees) | **13/20 = 65%** |
| Lane 3 | 14 pass / 6 fail — **30% failure**, dominating |
| Lanes 1 / 2 / 4 | 20/0, 19/1, 19/1 |

Gating today would have reddened roughly **one main run in three** for reasons unrelated to the change under test.

The trend does look better — older 10 runs 4/10 green, newer 10 9/10, and the last **7 consecutive** runs green. But no Windows fix merged in that window (I checked every commit; #2139 is still open), and at the recent failure rate a 7-green streak is `0.9^7 ≈ 48%` — about a coin flip. A streak with no causal change behind it is luck on a flaky lane, not stability.

**#2139 is the precondition and it is ready.** The most recent Windows failure on `main` was `star-map-composer-attachments.spec.ts` at 118/125/166/231/284, plus `star-map-project-pagination.spec.ts:9` and `star-map-chat-card-history.spec.ts:100` — exactly the Star Map readiness specs #2139 fixes. Its own CI has **all four Windows lanes green, lane 3 included**, at `MERGEABLE/CLEAN`.

So: merge #2139, let `main` hold a green Windows baseline, then promote the lane. The job comment now records the measurement and names that precondition, so the next person does not have to re-derive it.

PR #2175, opened 2026-09-15
fix(desktop): limit federation streams to visible thread views

huntharo:
## Summary

- The main window previously kept transcript streams subscribed for its five most recently selected remote threads. Switching from A to B now releases A's streaming demand while retaining its cached transcript.
- Reopening a cached remote thread requests catch-up using its owner revision, even when navigation timestamps have not changed or another window keeps the process-wide subscription alive.
- Replace cache-based subscription input with explicit additional visible thread panes. Star Map continues to support multiple open chat cards.

## Verification

- 175 tests passed across the thread-session, recent-remote-thread, Federation subscription, and Star Map subscription suites.
- `pnpm typecheck` passed.
- `pnpm lint:eslint` passed (warnings only).
- `git diff --check` passed.

## Notes

- Open Star Map chat cards still count as visible when panned off-screen. Keyboard focus does not control subscriptions.
- Streaming chunk cadence is unchanged. This change reduces which threads request transcript streams.


huntharo:
Not doing this... doesn't seem to be necessary right now

PR #2185, opened 2026-09-16
fix(messaging): search durable default-route destinations

huntharo:
Messaging Settings mixed channels and DMs with every observed reply thread in a
native dropdown, with no way to search and no way to tell two similarly named
destinations apart. This replaces the default-route surface selector with a
searchable picker built from the composer's branch/project picker primitives.

## What changed

**Durable destinations only.** Ephemeral reply threads are excluded from both
the observed choices and manual type selection — a reply thread belongs to a
binding, not to a default route. Telegram's named forum topics stay selectable
and get their own section.

**Search matches names and IDs.** Rows carry the destination's durable ID in a
right-aligned mono column, so `#orchard-planning` in two workspaces are
distinguishable at a glance rather than by guesswork.

**Grouped, not filtered.** Sections (`Current configuration`, `Channels &
groups`, `Direct messages`, `Telegram topics`) carry kind, so there is no
filter control to operate, nothing is hidden from view, and no control appears
and disappears depending on the platform.

**Keyboard and dismissal.** Arrow keys, Enter, Escape, outside dismissal, focus
restoration, and manual IDs. The list is bounded and scrollable, and the panel
expands inline within Settings — `.settings-panel` is `overflow: hidden`, so a
floating popover would be clipped by the card it opens inside.

Existing saved routes remain editable and removable; channel-wide child
defaults and workspace/provider/profile scopes retain their routing behavior.
Automation settings are outside this change.

Related merged PR #1258 fixed exact Slack channel routing, but did not address
this selector's mixed list or search.

## Screenshot

All names, IDs, and records below are fictional. No supplied user screenshots
or live profile data were used. Rendered against the real `app.css`.

Searchable default-route picker with fictional channel data

A UX design review of the first pass on this branch, and what the second commit
changed in response, is in the comments.

## Validation

Renderer tests across settings and styles (471); workspace typecheck; ESLint
(0 errors, existing warnings); renderer color lint; dependency boundaries.


huntharo:
## UX design review

Reviewed against the surfaces this was meant to be based on — the composer's
branch and project pickers. The class names came across; the design mostly did
not. Fixed in cc49cf604.

### Before / after

**First pass on this branch** (e57710f7c):

Surface picker before

**Revised** (c83abbb56):

Surface picker after

### What was off, and why it matters

**1. Two controls for one field.** The trigger kept reading
`Choose a recently seen surface...` with a closed chevron while its own panel
was open below it, so the field looked unset while you were searching in it.
The panel now replaces the trigger in place; focus returns on Escape or
selection.

**2. Popover chrome on an inline block.** The panel carried a popover's border,
elevated fill and `--shadow-popover` while being `position: relative` — lit
like it floats, behaving like it doesn't. The inline anchor is the right call
(`.settings-panel` is `overflow: hidden`, so a floating popover gets clipped),
so the fix was to stop pretending: no shadow, and the focus treatment of the
`.settings-select` it stands in for.

**3. The filter bank outweighed the results.** Three 34px `.button--ghost`
pills were the loudest thing in the panel, and the set read as All / A / B —
"Channels & DMs" is silently a superset of its two neighbours. Replaced with
`__section` eyebrows per kind. Grouping never hides a row, needs no operating,
and doesn't appear-and-disappear across platforms the way a Telegram-only tab
does. If the observed-surface list ever runs long enough that grouping stops
fitting, the alternate in the canvas below rebuilds a filter from
`.settings-segmented` — but that's a filter's job, not this list's.

**4. The row didn't tell destinations apart visually.** The stated goal was
distinguishing similarly-named destinations, but the answer was a wrapping
subline (`Channel · ID C_ORCHARD · Seen Sep 16, 2026, 8:00 AM`). Rows now use
the shipped anatomy — check column, kind glyph, name, ID right-aligned in
`__row-path`'s dim mono, last-seen trailing — on one line. Kind moved to the
eyebrow, so it isn't repeated per row.

**5. Manual entry read as a result.** No separator, no `+`, no accent, and no
check column — so it sat 20px left of every row above it. Now
`__separator` + `__row--action` + `__plus`, as in the project picker.

### Reuse, not restatement

The shared picker block in `app.css` says to edit the look there and keep only
genuinely per-picker overrides local. Five primitives had been re-implemented:

| Was | Now |
|---|---|
| `.button--ghost` + local `[aria-pressed]` | section eyebrows (`__section`) |
| `__identity` + `<small>`, two wrapping lines | `__row-name` + `__row-path` |
| `.messaging-routes__empty` | `.project-picker__empty` |
| bare `__row` for manual entry | `__separator` + `__row--action` + `__plus` |
| local `.is-active` fill override | `.is-cursor` in the shared hover/active group |

`is-active` also meant the wrong thing: in the project picker it is the
*selected* row (accent text, no fill, "the fill is reserved for hover /
keyboard focus"). It was being used for the keyboard cursor, then given a fill
locally. Selection is `is-active` again; the cursor is `is-cursor`.

One genuine sizing bug surfaced from rendering against the real stylesheet:
`__row-name` caps at 160px, which is right inside a ~420px popover and far too
tight in a panel as wide as the Settings form — every name truncated with most
of the row empty beside it.

### Accessibility

The trigger's `aria-label="Messaging surface"` overrode its visible text, so a
screen reader announced the field and never the chosen destination. Now
`Surface: <label>`, matching `ProjectPicker`'s `Project: <label>`.

### Design canvas

Reference pickers, as-built, the fix, the filter alternate, and a row-anatomy
spec: https://claude.ai/artifact/AG7dN6ALFmmUSWbGRApT9T

---

All destinations shown are fictional. Renders are the real `app.css` against
the markup each version emits.

Validation: 471 renderer tests (settings + styles), workspace typecheck,
ESLint 0 errors, renderer color lint, dependency boundaries.

PR #2192, opened 2026-09-17
style(tests): match adjacent formatting in the terminal resize tests

huntharo:
Formatting only. The tests [#2168](https://github.com/pwrdrvr/PwrAgent/pull/2168) added are written far denser than the code around them, and this repository has no formatter that will ever normalize it — so it stays dense until someone reflows it by hand. This reflows those blocks into the shape the surrounding tests already use.

## What changed

`apps/desktop/src/main/__tests__/integrated-terminal-service.test.ts` — the two tests "deduplicates normalized PTY sizes across viewers…" and "does not remember a rejected PTY resize as applied":

- the inlined `loadNodePty: async () => ({ spawn: … })` object, twice, now matches the form used ~15 lines below and everywhere else in the file
- the 137-character `createOrAttach({ … }, fakeWebContents())` call, and the request literal it shares with the sibling test, go one field per line, like the "does not add duplicate destroyed listeners…" test just below
- the `resize` helper wraps at `=>`; the nested `toEqual` array goes one row per line; `mockImplementationOnce(() => { throw … })` gets a real block body

`apps/desktop/src/renderer/src/features/thread-detail/__tests__/IntegratedTerminal.test.tsx` — the three `act`/`requestAnimationFrame` tests and the `mockAnimationFrames` helper:

- each `<IntegratedTerminal />` render gets one prop per line, as all seven earlier renders in the file do
- `act(() => { a(); b(); })` bodies split into statements, matching the four existing `act(() => {` blocks
- the two `toHaveBeenCalledExactlyOnceWith` arguments wrap
- `mockAnimationFrames` returns a normal multiline literal with a trailing comma, and `cancelAnimationFrame`'s implementation gets a block body like `requestAnimationFrame`'s directly above it

## Effect on line length

The longest **code** line (excluding `it` titles, which are strings) in each file:

| File | Before | After | Longest pre-existing |
|---|---|---|---|
| `integrated-terminal-service.test.ts` | 137 | 88 | 107 |
| `IntegratedTerminal.test.tsx` | 122 | 79 | 89 |

## No behavior change

Verified mechanically, not by eye: both files were compared against `HEAD` with a lexer that keeps string literals byte-exact and preserves identifier boundaries. The token streams are identical apart from two sanctioned additions — the trailing commas the root `CLAUDE.md` requires on a multiline literal, and the parentheses that let the `view` arrow's element span lines, which is the shape the `view` arrow ~120 lines above already uses.

No renamed variables, no altered assertions, no added or removed cases. Nothing outside the listed blocks was touched. No Prettier and no `eslint --fix` — hand-formatted, per the root `CLAUDE.md`.

## Verification

- `pnpm test <both files>` — 57 passed
- `pnpm typecheck` — clean
- `pnpm lint:eslint` — exit 0 (69 pre-existing warnings, 0 errors)


huntharo:
Closing on review of the whole batch rather than of this diff.

This came from a `/code-review --fix` pass on an already-merged PR. In that review I ranked this finding last and described it as the first I would dismiss — then spawned it as a task anyway, which inverted my own judgment. Reformatting merged, passing tests buys nothing and cuts against the repo's "keep each diff limited to the requested change" rule.

The formatting observation stands if someone is editing those tests for another reason.

PR #2193, opened 2026-09-17
refactor(desktop): give the renderer one frame coalescer

huntharo:
## What

Adds `apps/desktop/src/renderer/src/lib/frame-coalescer.ts` — a plain
`createFrameCoalescer(run) → { schedule, cancel }` — and moves the renderer's
six hand-written "coalesce this onto the next frame" blocks onto it.

De-duplication only. No behavior change at any call site.

## Why

Each copy re-answered the same three questions from scratch:

1. Is a request that arrives while a frame is queued dropped, or does it replace the queued work?
2. Is the frame cancelled on teardown, and by whom?
3. Is the "should this run at all" guard checked at schedule time or at frame time?

All six agreed on (1) drop and (2) cancel, owned by whatever tears the
subject down. Nothing made them agree — the agreement was six independent
coincidences.

Dropping rather than replacing is not a loss: every one of these schedules a
closure that reads its inputs *at frame time* (the drag's latest travel, the
terminal's current grid), so the dropped requests still reach the one frame
that runs. There is a unit test pinning that specifically, because it is the
property that makes "drop" the correct answer rather than merely the cheap one.

**Question 3 stays at the call site on purpose.** A schedule-time guard and a
frame-time guard are different behaviors and the difference has to stay
visible where it is decided, so the primitive takes no predicate.
`IntegratedTerminal` means both of its guards and keeps both: it refuses to
schedule once `disposed`, and re-checks `visibleRef` *inside* `fitAndResize`,
so a fit queued across a visibility flip settles against the size the terminal
actually has. A primitive that hoisted that check to schedule time would have
broken it. `DirectoriesList` likewise keeps its `activated`/`canceled` test
before `schedule()`.

## Call sites moved

| Call site | Work coalesced |
|---|---|
| `IntegratedTerminal.tsx` | `scheduleFitAndResize` (#2168) |
| `StarMapScreen.tsx` | canvas pan drag's clamped repaint |
| `useStarMapCardDrag.ts` | card drag's DOM write + group delta |
| `App.tsx` | sidebar resize's `--sidebar-width` flush |
| `SettingsScreen.tsx` | document scroll clamp |
| `DirectoriesList.tsx` | thread-pin drag's drop-target hit test |

The task named the first two; the remaining four came out of a grep of the
whole renderer. All six fit without an option flag, so none were left behind.

Two small knock-on cleanups fell out: `App.tsx`'s `flush` no longer clears the
frame id as a side effect of doing its work (`stop` calls it directly after
`cancel`), and `SettingsScreen`'s component-level `scrollClampFrameRef` is gone
— it only ever held a frame id for one effect and its cleanup.

## Deliberately not moved

The grep also turns up two other `requestAnimationFrame` shapes that are not
this one:

- **Self-rescheduling animation loops** — `useStarMapFlight`,
  `useStarMapCameraKeys`, `ThreadFindBar`'s scroll-settle tick. These run
  every frame until a condition ends them. `useStarMapCameraKeys`'s
  `if (frame) return` reads like a coalescing guard but is a loop-start guard:
  `runFrame` reschedules unconditionally, so the handle is never clear while
  flying.
- **One-shot "next frame" deferrals** — the focus-after-paint calls in
  `Composer`, `ReferencePicker`, `ThreadRow`, and the double-rAF waits in
  `ThreadView` and `App`. Nothing repeats, so there is nothing to coalesce.

Forcing either onto this primitive is how it would grow the booleans.

## Verification

- `pnpm test apps/desktop/src/renderer` — 317 files, 4724 tests, all pass
- `pnpm typecheck`, `pnpm lint:eslint` (0 errors), `pnpm lint:boundaries` (no violations) — all clean

The strongest evidence for "semantics preserved" is that the existing tests
did not need touching. `IntegratedTerminal.test.tsx` already pins all three
answers for that call site — "coalesces terminal fitting outside observer
delivery", "cancels a queued fit when the terminal unmounts", and "skips a
queued fit while hidden and fits the current grid when shown" — driving real
frames through a `window.requestAnimationFrame` spy. They pass unchanged, as
does the Star Map drag test that awaits a real frame. The new unit test covers
the primitive's own edges: cancel-with-nothing-queued, reuse after cancel,
self-rescheduling from inside a frame, and independence between coalescers.

No user-visible UI change, so no screenshots.


huntharo:
Closing on review of the whole batch rather than of this diff.

This came from a `/code-review --fix` pass on an already-merged PR, as a 'reuse' finding: four hand-written animation-frame coalescers with no shared primitive. That observation is accurate, but it does not justify what it turned into — a ~370-line refactor of working code, with no behavior change, landing immediately after separate bug fixes in one of the same files (#2190).

The duplication is worth folding in the next time one of those call sites is edited for a real reason, not as its own change.

PR #2196, opened 2026-09-17
fix(messaging): support automation triggers and results across provider surfaces

huntharo:
## What changes

Selecting a DM contact previously saved a channel-shaped user ID. Slack, Discord, and Mattermost inbound DMs use native conversation IDs instead, and result delivery could post to the wrong kind of identifier. Automations also missed unmentioned Slack group-DM and LINE group/room messages before they reached the matcher.

This change carries explicit contact-recipient identity through the editor and stored automation, matches only 1:1 DM events for a contact, and resolves recipients inside each provider before `messaging_target` delivery. Native group-DM and room destinations retain their conversation IDs. Matching and preview now share provider/kind/parent-conversation checks, including Telegram topic parent scoping and Discord channel ancestry. Mattermost DM replies retain DM authorization semantics. Discord forwards other authorized bots while suppressing itself.

Authorized ambient Slack group-DM and LINE group/room traffic is observation-only: it reaches preview and matching automations, without becoming ordinary Agent input. Existing authorization gates remain in place.

History capability is reported by the main process. Only Slack currently implements native top-level conversation history; contact-based DM and scoped thread/topic previews explicitly use live capture without promising history/replay. Prompts identify DM sources even without a title.

The minimal editor wiring includes all six providers' configured conversations and contacts because #2186 was not present in the fetched `main`. There is no layout redesign.

## Verified surface matrix

| Provider | Inbound and output surfaces covered by adapter code/tests |
| --- | --- |
| Telegram | Private chats, authorized groups/supergroups, forum topics |
| Discord | 1:1 DMs, guild channels and threads; user-to-DM resolution |
| Slack | 1:1 DMs and replies, channels/threads, policy-authorized group DMs |
| Mattermost | 1:1 and group DMs, replies, channels; user-to-DM resolution |
| Feishu/Lark | p2p and group chats; `open_id` versus `chat_id` delivery |
| LINE | User chats, groups, and rooms; unmentioned shared traffic and push targets |

All verification uses real adapters with fake SDK/API seams, plus main-process matcher, preview, controller, executor, and renderer tests. **No live accounts were exercised.** Account scopes, membership, webhook subscriptions, privacy settings, blocked/unstarted DMs, and server permissions remain live-verification requirements.

Genuinely unsupported here: Discord bot group DMs; Telegram broadcast `channel_post` ingress; history in the five non-Slack adapters; history/replay for contact-addressed DMs and scoped threads/topics. Feishu threaded messages remain chat-level; Mattermost retains its webhook anti-echo exclusion. DM code capture still directs operators to authorized-contact selection rather than approving a user as a group.

Detailed evidence and limitations: [automation messaging surface verification](docs/automation-messaging-surfaces.md).

## Validation

- `pnpm test automation inbound-preview packages/messaging apps/desktop/src/main/__tests__/messaging apps/desktop/src/main/__tests__/telegram-adapter.test.ts apps/desktop/src/main/__tests__/discord-adapter.test.ts` — **103 files, 1,757 tests passed**.
- `pnpm lint` — passed, including SQL/storage/color/license checks, ESLint, typecheck, and dependency boundaries. ESLint reports existing warnings and zero errors.
- Final `pnpm typecheck` and `pnpm lint:eslint` rerun after the last small edits.
- `git diff --check` — passed.

An earlier run hit the existing cold-import runtime timeout while lint/typecheck competed for CPU. The final full test command passed without changing timeouts, workers, or test isolation.

No SQLite schema, write call, timer, or background persistence was added. Accepted traffic uses the existing messaging and automation persistence paths.


huntharo:
## UX design review

Full write-up: [`docs/design/automation-messaging-surfaces-ux-review.md`](docs/design/automation-messaging-surfaces-ux-review.md).
Before/after prototype: [`docs/design/automation-surfaces-v1/`](docs/design/automation-surfaces-v1/) (open `index.html`, append `?slide=<id>` to isolate a sheet).

The plumbing here is sound and `docs/automation-messaging-surfaces.md` is unusually honest about its own limits. This review covers the half the handoff deferred: whether an operator can now **tell** what the plumbing will do.

Short version — the feature reads as finished from the main process outward and unfinished from the operator inward. DMs became selectable; the vocabulary around them did not change. All fixture data below is fully contrived.

### 🔴 1 — The `dm:` sentinel reaches the picker's durable-ID column

`readProviderGroups` encodes contacts as `` `dm:${contact.id}` ``, and `conversationOptions` derives `detail` from that same value. The mono ID column — whose own docstring says it exists so that two destinations sharing a name "are told apart by that ID column, which is what the native `<select>` it replaced could not do" — now prints an ID that exists nowhere on the platform, and that the search box matches against. A contact with no display name prints the same identifier twice, spelled two ways.

surface picker before and after

### 🔴 2 — Manual entry cannot express a DM, and silently saves an inert trigger

`buildDestinationSnapshot` only produces a DM for a `dm:`-prefixed value, which only the picker can produce. Anything typed manually saves `conversationKind: "channel"`.

Verified against the shipped matcher: a `channel`-kind target is rejected against a real DM on **Slack, Mattermost, Feishu and Discord**; **Telegram and LINE** accidentally succeed via this PR's legacy-ID fallback. Same operator action, two outcomes, no feedback in either case. The Slack hint makes it likely — "open the channel details and copy the ID at the bottom" yields the `D…` DM conversation ID.

This is the failure mode the handoff set out to eliminate. It was removed from the picker path and left in place on the manual path.

manual entry before and after

### 🔴 3 — The replay empty state blames the provider for a scope limit

This PR added two refusals to `listReplayCandidates` (`recipientUserId`, `parentId`). `AutomationsScreen` still renders the sentence written when the provider was the only refusal. For a Slack DM trigger, *"This provider can't serve conversation history"* is false — the same automation would offer replay on a channel.

replay empty state before and after

### 🟡 4 — A live Telegram destination field whose value is discarded

The trigger side suppresses topic controls for a DM; the destination side guards on `destProvider === "telegram"` alone, while `buildDestinationSnapshot` returns before reading `topicId`. The field stays on screen for a 1:1 DM, accepts input, and drops it on save.

destination before and after

### 🟡 5–6 — Nouns that promise what cannot appear, and three providers with no guidance

Discord's only authorized list is of guilds, so the picker offers DMs and nothing else — while the field says *Channel*, the trigger says *"Choose a channel or DM"*, and the *"Authorized channels"* heading never renders. Separately, Mattermost, Feishu and LINE — the three providers this PR just made selectable — fall through to *"The conversation ID PwrAgent should watch."*

discord and generic providers before and after

### 🟡 7–8 — Copy that describes a room, and a limit stated too late

`every message in Dana Okonkwo` — and it is also imprecise: the matcher requires `actorId === recipientUserId`, so a DM trigger fires on messages **from** that person. The preview panel calls the trigger source a *"destination"*, which is the name of a different field later in the same form; `previewHistorySupported` starts `false`, so it asserts "History is unavailable" for a frame before the IPC answers; and `supportsPreviewHistory` is **synchronous**, so this could be said under the button rather than after the panel opens.

preview before and after

### What works well

- **One matcher, three surfaces.** `matchesAutomationConversation` is now shared by the live matcher, the preview and the replay badges, so the "matches" badge is a real promise about what the trigger would do.
- **"Authorized …" headings** explain a short list instead of leaving it looking broken.
- **The kind glyph column** does the work a kind filter would have done without hiding rows, and survives a saved route whose kind is no longer selectable.
- **No contrast failures.** `--text-muted` on `--bg-panel` is 5.4:1; rows are 38px, fields 32px.

Fixes for 1–4 are going up next on this branch.

huntharo:
## Findings 1–4 are fixed on this branch

`fad4e46b`. Gates: `typecheck`, `lint:eslint` (0 errors), `lint:boundaries`, and the automations + messaging + shared suites — 1,636 tests, all passing. 9 new tests cover the four fixes, including the three refusal reasons and the "older build sent no reason" fallback.

The prototype sheets are updated: a **green** right-hand column is what the branch does now, an **orange** one is still a proposal. Four sheets carry both.

**1 — the `dm:` marker is out of the ID column.** One `contactUserId` helper decides what `dm:` means; `conversationOptions` shows the platform ID, or nothing when it would repeat the label.

**2 — manual entry can say "this is a DM".** A Channel/Direct message switch on both manual fields, writing the same `dm:` form the picker produces so nothing downstream learns a second encoding. Label, placeholder and hint follow the switch; the DM hint asks for the sender's user ID — what `matchesAutomationConversation` actually compares — and names the conversation ID it is *not*, per provider. Toggling re-encodes what is already typed rather than clearing it.

**3 — replay names the constraint that applied.** `ListAutomationReplayCandidatesResponse` now carries `unsupportedReason` (`contact_dm` / `scoped_thread` / `provider`) beside `supported`, checked most-specific-first so a Slack DM reports the scope. Four sentences: one per reason, plus a fallback that blames nothing when the field is absent.

**4 — the Telegram destination topic field is gone for a DM**, via the same `dm:` test the trigger side already uses.

surface picker before and after

manual entry before and after

replay empty state before and after

destination before and after

### Still open — findings 5–9

A copy pass, no structural change. Drafted in the sheets, not implemented:

- **5** Discord's picker offers DMs only (guilds are servers), while every noun around it says *channel* and the "Authorized channels" heading never renders. Worth doing next — an operator is currently told to look for a section that cannot exist.
- **6** Mattermost, Feishu and LINE still share `"The conversation ID PwrAgent should watch."` on the *channel* path. Their DM path got per-provider copy in this commit; the channel path did not.
- **7** `every message in Dana Okonkwo`, and the picker's `aria-label` composing `"Destination group: Dana Okonkwo"`.
- **8** The preview panel calls the trigger source a *destination*; `previewHistorySupported` starts `false` so it asserts "History is unavailable" for a frame; and `supportsPreviewHistory` is synchronous, so this could be said under the button instead of after the panel opens.
- **9** "Include thread replies" on 1:1 DM triggers; capture-by-code saying "channel" for three providers; "Messaging settings" vs "Settings > Messaging"; and `previewScope`'s object literal indented at the wrong level for its `? :`.

Happy to take any of these in a follow-up — say the word.

